PR #4, opened 2019-04-04
Ether HTLC with dynamic final address

### LINKED ISSUE #40: ⚗️ Can RFC7 Ether HTLC be specified without knowing identities upfront
The Bitcoin and Ether HTLCs have different functionality. The Bitcoin HTLC lets you specify which identity will finally own the funds upon *redeeming* while the Ether HTLC makes you specify the final address at contract creation time.

By restructuring the Ether HTLC we may be able to give it this functionality and therefore ease the burden on implementations.

LLFourn:
resolves https://github.com/comit-network/RFCs/issues/40

Note I restricted the scope of this spike to not include EVM assmebly. We can figure out the actual assembly after we've done https://github.com/comit-network/comit-rs/issues/606

### DIFF
--- a/0002-ether-htlc-dynamic-final-addresses.md
+++ b/0002-ether-htlc-dynamic-final-addresses.md
@@ -76,7 +76,7 @@ To verify the signature we use the `ecrecover` precompiled contract.
 Note that this means the signature we need to send is the 65 byte recoverable ECDSA signature.
 (A recoverable signature is one where you can figure out the public key from the signature itself)
 
-Here's the psuedocode for the contract:
+Here's the rust-styled pseudocode for the contract:
 
 ``` rust
 let (expiry, redeem_identity, refund_identity, secret_hash) = embedded_somehow...;
--- a/0002-ether-htlc-dynamic-final-addresses.md
+++ b/0002-ether-htlc-dynamic-final-addresses.md
@@ -58,12 +58,15 @@ The key things when deciding what to sign are:
 2. It must not be possible to re-use the signature on a different chain or a different contract on the same chain.
 
 To ensure (1) we need to sign the final address (redeem or refund) with the private key of the corresponding identity.
-To ensure (2) we should sign the HTLC's contract address, the network id and chain id.
+To ensure (2) we should sign the HTLC's contract address, the chain_id
+
+Note, that hashing the `network_id` isn't necessary because it doesn't seem to be used for protecting against replay attacks in Etheruem.
+No current Ethereum chain has the same `chain_id` (see [chainid.network/](https://chainid.network/)), regardless of their `network_id` so they can be uniquely identified by their `chain_id` alone.
 
 So I propose that we sign the following messages for each outcome:
 
-- Redeem: `network_id || chain_id || htlc_contract_address || final_redeem_address`
-- Refund: `network_id || chain_id || htlc_contract_address || final_refund_address`
+- Redeem: `chain_id || htlc_contract_address || final_redeem_address`
+- Refund: `chain_id || htlc_contract_address || final_refund_address`
 
 We have to hash the message before signing it.
 In this case we can choose any 256 bit output hash function because we control both the singing code (in COMIT) and the verifier (in the contract).
@@ -86,7 +89,7 @@ if CALLDATASIZE == (20 + 65 + 32) {
     let (final_redeem_address, signature, secret) = CALLDATA();
 
     if sha256(secret) == secret_hash {
-        let message_hash = sha3(network_id, chain_id, address(), final_redeem_address);
+        let message_hash = sha3(chain_id, address(), final_redeem_address);
         let (v,r,s) = signature;
         if redeem_identity == ecrecover(message_hash, v, r, s) {
             log("Redeemed(secret)"); // could also log the address it was redeemed to
--- a/0002-ether-htlc-dynamic-final-addresses.md
+++ b/0002-ether-htlc-dynamic-final-addresses.md
@@ -90,7 +90,7 @@ if CALLDATASIZE == (20 + 65 + 32) {
         let (v,r,s) = signature;
         if redeem_identity == ecrecover(message_hash, v, r, s) {
             log("Redeemed(secret)"); // could also log the address it was redeemed to
-            selfdescrtruct(final_redeem_address)
+            selfdestruct(final_redeem_address)
         }
     }
     else {
@@ -101,7 +101,7 @@ if CALLDATASIZE == (20 + 65 + 32) {
 
     if timestamp() > expiry {
         log("Refunded");
-        selfdescrtruct(final_refund_address);
+        selfdestruct(final_refund_address);
     }
     else {
         return
--- a/0002-ether-htlc-dynamic-final-addresses.md
+++ b/0002-ether-htlc-dynamic-final-addresses.md
@@ -1,0 +1,121 @@
+# Ether HTLC with dynamic final addresses
+
+* Authors: Lloyd Fournier
+* Date: 2019-04-04
+
+Issue: [#40](https://github.com/comit-network/RFCs/issues/40)
+
+## Context
+
+The RFC003 [Bitcoin HTLC](https://github.com/comit-network/RFCs/blob/master/RFC-005-SWAP-Basic-Bitcoin.md) and [Ether HTLC](https://github.com/comit-network/RFCs/blob/master/RFC-007-SWAP-Basic-Ether.md) have different functionality.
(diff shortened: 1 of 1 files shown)


0002-ether-htlc-dynamic-final-addresses.md
```diff
@@ -0,0 +1,121 @@
...
+
+The decision we have to make here is whether the final address is:
+
+1. Provided in the data
+2. The sender address
+
+(2) is nice because it means that there is no way to mess it up and send a transaction to an address you don't own. It also saves a bit of gas.
+
+On the other hand, that might be a feature.
+You may want to send the asset in the HTLC to a contract or something, so I'm leaning towards (1).
```
D4nte, line 50: I am more keen on (2) for simplicity/cost/security (as you mention) reasons and change to (1) if the feature is requested.

bonomat: I think if we don't want to change our core concepts then (1) is the way to go: 

* If we go for (2) the comit-node would need funds to send a transaction from the intermediate account (if we want to use the same approach as for Bitcoin).
* You mentioned we want to mimic Bitcoin

0002-ether-htlc-dynamic-final-addresses.md
```diff
@@ -0,0 +1,121 @@
...
+1. It must not be possible to front run the transaction, learn the secret, and then create another transaction that redeems to a different address.
+2. It must not be possible to re-use the signature on a different chain or a different contract on the same chain.
+
+To ensure (1) we need to sign the final address (redeem or refund) with the private key of the corresponding identity.
+To ensure (2) we should sign the HTLC's contract address, the network id and chain id.
+
+So I propose that we sign the following messages for each outcome:
+
+- Redeem: `network_id || chain_id || htlc_contract_address || final_redeem_address`
+- Refund: `network_id || chain_id || htlc_contract_address || final_refund_address`
```
bonomat: I shortly mentioned it on the ticket but let's discuss here: 

We will need to be able generate the same message inside the contract (and hash it) as as outside of the contract, otherwise the caller can pass an arbitrary hash. 

Since there is no opcode for retrieving the `network_id` or `chain_id` we would need to hardcode it into the contract. This would make using these values in the message for a replay attack (in case of a fork) rather useless, as they will always be the same. 

Instead, we should limit the message to be signed to something which is unique to the current swap, e.g. a combination of the `hashed_secret` + the `contract_address` and require the message to be signed from the `redeem_identity` (or `refund_identity` respectively). 

This would not protect us from a replay attack in case of a fork (I probably wouldn't call it attack, it's just a replay with the same behavior), but then again, we want to mimic Bitcoin and depending on the fork, I'd expect the same behavior.

0002-ether-htlc-dynamic-final-addresses.md
```diff
@@ -0,0 +1,121 @@
...
+
+We have to hash the message before signing it.
+In this case we can choose any 256 bit output hash function because we control both the singing code (in COMIT) and the verifier (in the contract).
+Our choices are `SHA-256` and `Keccak-256`.
```
bonomat, line 73: The reason why I mentioned sha-256 was actually because I wasn't sure of which hash function `ecrecover` would accept. All I could find was that it's a hashed messaged :) Did you verify that? Otherwise we can just drop sha-256.

0002-ether-htlc-dynamic-final-addresses.md
```diff
@@ -0,0 +1,121 @@
...
+
+// We're redeemimg with secret
+if CALLDATASIZE == (20 + 65 + 32) {
+    let (final_redeem_address, signature, secret) = CALLDATA();
+
+    if sha256(secret) == secret_hash {
+        let message_hash = sha3(network_id, chain_id, address(), final_redeem_address);
+        let (v,r,s) = signature;
+        if redeem_identity == ecrecover(message_hash, v, r, s) {
+            log("Redeemed(secret)"); // could also log the address it was redeemed to
```
bonomat, line 95: `log("Redeemed(secret)");` is correct as we need to print the secret to extract it easily from the transaction receipt.

0002-ether-htlc-dynamic-final-addresses.md
```diff
@@ -0,0 +1,121 @@
...
+Our choices are `SHA-256` and `Keccak-256`.
+I think we should just use `Keccak-256` because it's less gas on EVM.
+
+### Verifying the signature
+
+To verify the signature we use the `ecrecover` precompiled contract.
+Note that this means the signature we need to send is the 65 byte recoverable ECDSA signature.
+(A recoverable signature is one where you can figure out the public key from the signature itself)
+
+Here's the psuedocode for the contract:
```
bonomat: rust-styled pseudo code 👏❤

0002-ether-htlc-dynamic-final-addresses.md
thomaseizinger: I would vote for option 1 mainly because of the reasons Philipp mentioned.

In addition, I see the following problem:

Sending the money to the sender's address requires the user to choose, which address the want to send it from when fetching the action. As you mentioned further down, the final redeem address needs to be signed for security reasons.

What if the user sends it from a different address though? This would mean the redeeming fails and debugging that requires serious knowledge about how the contracts work. To avoid this problem altogether, we should put the final redeem address into the `data`. This allows the redeeming to happen from any address and not just the address the user wants the money on.

I think (2) can cause way too many problems here and the overhead of (1) seems to be negligible (only the cost of the added bytes in the TX I presume).

0002-ether-htlc-dynamic-final-addresses.md
thomaseizinger:
```suggestion
Here's the pseudocode for the contract:
```

0002-ether-htlc-dynamic-final-addresses.md
```diff
@@ -0,0 +1,121 @@
...
+// We're redeemimg with secret
+if CALLDATASIZE == (20 + 65 + 32) {
+    let (final_redeem_address, signature, secret) = CALLDATA();
+
+    if sha256(secret) == secret_hash {
+        let message_hash = sha3(network_id, chain_id, address(), final_redeem_address);
+        let (v,r,s) = signature;
+        if redeem_identity == ecrecover(message_hash, v, r, s) {
+            log("Redeemed(secret)"); // could also log the address it was redeemed to
+            selfdescrtruct(final_redeem_address)
```
thomaseizinger:
```suggestion
            selfdestruct(final_redeem_address)
```

0002-ether-htlc-dynamic-final-addresses.md
```diff
@@ -0,0 +1,121 @@
...
+    }
+    else {
+        return
+    }
+} else {
+    let (final_refund_address, signature) = CALLDATA();
+
+    if timestamp() > expiry {
+        log("Refunded");
+        selfdescrtruct(final_refund_address);
```
thomaseizinger:
```suggestion
        selfdestruct(final_refund_address);
```

0002-ether-htlc-dynamic-final-addresses.md
LLFourn: IMO the best course of action when making these decisions is to simply guarantee that the TX can never be reused. For that you need the `network_id` and `chain_id` embedded in the contract.

Like you said, you could sign the `secret_hash` instead of these so a replay attack would only work against the same contract on a different chain if it has the same secret AND the same redeem or refund identity (which would be pretty strange). But rather than thinking about how we could justify this never happening, my preference would be to make it impossible.

0002-ether-htlc-dynamic-final-addresses.md
LLFourn: `ecrecover` doesn't care. It just takes 32 bytes (within the secp256k1 curve order I think). Keccak-256 it is then.

0002-ether-htlc-dynamic-final-addresses.md
LLFourn: Yeah I wasn't sure if we cared about the address it was redeemed to in terms of transparency. With Bitcoin we can obviously tell where the redeem went to, if we want to be able to do the same here we need to log the address in the data too.
I'm in favour of logging the address as well.

D4nte (approved):

LLFourn:
Let's discuss together on Wed when we no longer have planning but still have a room booked

LLFourn:
I did some further research and removed the `network_id` from the hash (`chain_id` is enough).

PR #5, opened 2019-04-16
Add Secret seed storage and key derivation spike

### LINKED ISSUE #672: ⚗️Key derivation and secret seed storage
The purpose of this spike is answer the following:

How should we take some secret data and use it to derive:

- transient keys within the various protocols 
- the node identifying key

How should we store this secret?
How should the generation happen? Automated or Manual?
Should we provide backup strategies (ie, mnemonic)

hint: look at other projects an summarise the best approaches

Child of #744

LLFourn:
resolves https://github.com/comit-network/comit-rs/issues/672

Note I only looked at lnd for inspiration. I couldn't find many other projects that did anything notable.

### DIFF
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -157,7 +157,7 @@ It faces a similar problem to us in that you can't restore stuff you've learned 
 
 The wallet seed is stored encrypted with a password in the node's `wallet.db` (or at least I think that's where it's hiding).
 You have to unlock the wallet on lnd startup with the password via an RPC call (this is what the ln-cli does).
-There is a (to be depreciated) `--noseedbackup` option  which encrypts it with the default password so you don't have unlock the wallet.
+There is a (to be deprecated) `--noseedbackup` option  which encrypts it with the default password so you don't have unlock the wallet.
 But this means you can't back up your seed so it's strictly for development.
 
 Here's an interesting issue which discusses the possibility of making the password optional  (i.e. leaving the seed unencrypted): [#899](https://github.com/lightningnetwork/lnd/issues/899) (note they removed the `--noencryptwallet` thing mentioned there).
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -51,13 +51,7 @@ This lets users backup the secret seed separately from the database (although th
 Given that the two are linked, we should store a hash of the secret seed along with the swap id (in the same row or whatever).
 This allows the user to change the secret seed while not invalidating the entire database.
 
-By default I'd put it somewhere like:
-
-```
-~/.comit/secret_seed
-```
-
-But we could also use something like https://github.com/soc/directories-rs to choose the directory for us.
+By default I'd put it next to the config file.
 
 But its location should be configurable with `secret_seed_file` in the `.toml` config.
 
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -143,8 +143,12 @@ The implementation I recommend using is https://github.com/RustCrypto/hashes jus
 
 ### NodeID
 
+This is the NodeID that will be used by libp2p.
+
 I think we should derive the NodeID from the master seed simply by hashing something arbitrary.
 We should leave the door open to having multiple top level ids (i.e. one for lightning, one for COMIT).
+
+The derivation path can be: `master_seed / "comit_node_id"`
 
 # Research Notes
 
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -149,7 +149,7 @@ We should leave the door open to having multiple top level ids (i.e. one for lig
 To help inform the above decisions I looked at LND because it's the flagship layer two protocol node and has to generate many keys and secrets.
 It faces a similar problem to us in that you can't restore stuff you've learned from the other party (signatures on states, revocation keys etc) from the seed.
 
-### Seed srorage
+### Seed storage
 
 The wallet seed is stored encrypted with a password in the node's `wallet.db` (or at least I think that's where it's hiding).
 You have to unlock the wallet on lnd startup with the password via an RPC call (this is what the ln-cli does).
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -68,7 +68,7 @@ We *could* encrypt the database with the `secret_seed` seeing as they're separat
 
 ### How to generate it?
 
-If `secret_seed_file` isn't set, I think we should just generate it from `OsRng` to `~/.comit/secret_seed` if it doesn't exist.
+If `secret_seed_file` isn't set, I think we should just generate it from `OsRng` to its default location if it doesn't exist.
 
 We can make a tool to generate/restore it from a mnemonic and encrypt/decrypt it later on.
 
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -76,6 +76,10 @@ If `secret_seed_file` isn't set, I think we should just generate it from `OsRng`
 
 We can make a tool to generate/restore it from a mnemonic and encrypt/decrypt it later on.
 
+### Should it have a checksum?
+
+This probably won't catch much but it seems like a decent idea to attach a checksum to it if it's stored or transferred in a way that could be corrupted (or perhaps corrupted accidentally via a text editor).
+
 ## Secret/NodeID Derivation
 
 I think the way we are doing it now is roughly the right way but I recommend modifying it in the following ways:
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -16,7 +16,7 @@ I present the conclusions first and a summary of what I learned from LND.
 
 This section recommends how we should generate and store the secret seed by answering key questions.
 
-### Is there any way to recover anything if you have your secret seed but no database?
+### Can you recover anything if you have your secret seed but no database backup?
 
 **In practice, No.**
 
@@ -30,24 +30,24 @@ I don't think this partial recovery is worth putting the significant effort that
 ### Should it be encrypted?
 
 I think this should be optional.
-A COMIT node doesn't own the funds long term.
-It can only control the funds of protocols that are in being executed.
+A COMIT node doesn't hold funds long term.
+It can only control the funds during the execution of a protocol.
 This is different to LND where if someone gets access to your seed they can take all your funds regardless of whether you have channels open or not.
 
-The best attack vector against an unencrypted seed is for the attacker to get short term access to the system COMIT is running on, covertly steal their secret seed and then wait for them to execute a swap and then redeem to your own address.
+The best attack vector against an unencrypted seed is for the attacker to get short term access to the system COMIT is running on, covertly steal their secret seed and then wait for them to execute a swap and then redeem to their own address.
 Encrypting the seed does give you protection against this.
 
 Having said that, giving an attacker access to your COMIT node will probably give them access to the HTTP API which they can use to redeem funds to their own address.
 This is probably a better attack vector than the above.
 
-note: We could prevent this attack by giving the node a way of generating the final redeem addresses without getting them from the api.
+Note: We could prevent this attack by giving the node a way of generating the final redeem addresses without getting them from the api.
 
-I think we should leave it unencrypted for now, make encryption an option later on and then perhaps make encryption the default as a last step.
+For the sake of expediency I propose we incrementally increase the security, i.e. (i) leave the secret unencrypted for now, (ii) make encryption an option later on and (iii) make encryption the default eventually.
 
 ### Where to store it?
 
-Unlike lnd, I don't think we should store secret seed in the database file.
-This allows users to backup the secret seed separately from the database (although they need to backup both).
+I don't think we should store secret seed in the database file (this is what lnd does).
+This lets users backup the secret seed separately from the database (although they need to backup both).
 Given that the two are linked, we should store a hash of the secret seed along with the swap id (in the same row or whatever).
 This allows the user to change the secret seed while not invalidating the entire database.
 
@@ -68,7 +68,7 @@ We should make the permissions restrictive so only the user has read permission 
 I don't think it's worth encrypting the database yet until we have a motivation to do it.
 An attacker stealing the database would learn what protocol executions you're involved in so atm it's just a privacy issue.
 
-We *could* encrypt the database we the `secret_seed` seeing as they're separated.
+We *could* encrypt the database with the `secret_seed` seeing as they're separated.
 
 ### How to generate it?
 
@@ -97,14 +97,14 @@ Obviously, in our case you can't do this attack but I'd suggest we use a hash fu
 ### Use u8 rather than strings to identify key types
 
 There's nothing inherently wrong with using utf-8 encoded strings to a hash function to produce secret keys etc. e.g. `hash(seed || "secret")` (putting length extension attacks to the side).
-This is what we do now in rfc003's [SecretSource](https://github.com/comit-network/comit-rs/blob/master/application/comit_node/src/swap_protocols/rfc003/secret_source.rs).
+This is what we do now in rfc003's [SecretSource](https://github.com/comit-network/comit-rs/blob/48d807d3b173f62b3e6adb2c5aebcb2ebb6967a5/application/comit_node/src/swap_protocols/rfc003/secret_source.rs).
 There is another way to do it which is to simply encode integers for each type of key:
 
 ``` rust
 enum SecretType {
-    Secret,
-    RedeemKey,
-    RefundKey,
+    Secret = 0,
+    RedeemKey = 1,
+    RefundKey = 2,
 }
 
 ..
@@ -135,6 +135,8 @@ The main reasons are:
 
 BLAKE2b's key is 64-bytes so I also suggest changing the secret seed we use to 64-bytes.
 
+The implementation I recommend using is https://github.com/RustCrypto/hashes just because it's maintained by the rust crypto organisation.
+
 ### NodeID
 
 I think we should derive the NodeID from the master seed simply by hashing something arbitrary.
@@ -151,7 +153,7 @@ It faces a similar problem to us in that you can't restore stuff you've learned 
 
 The wallet seed is stored encrypted with a password in the node's `wallet.db` (or at least I think that's where it's hiding).
 You have to unlock the wallet on lnd startup with the password via an RPC call (this is what the ln-cli does).
-There is a `--noseedbackup` option which id going to be depreciated which encrypts it with the default password so you don't have unlock the wallet.
+There is a (to be depreciated) `--noseedbackup` option  which encrypts it with the default password so you don't have unlock the wallet.
 But this means you can't back up your seed so it's strictly for development.
 
 Here's an interesting issue which discusses the possibility of making the password optional  (i.e. leaving the seed unencrypted): [#899](https://github.com/lightningnetwork/lnd/issues/899) (note they removed the `--noencryptwallet` thing mentioned there).
--- a/0003-secret-seed-storage-and-key-derivation.md
+++ b/0003-secret-seed-storage-and-key-derivation.md
@@ -1,0 +1,195 @@
+# Secret seed storage and key derivation
+
+* Authors: Lloyd Fournier
+* Date: 2019-04-16
+
+Issue: [#672](https://github.com/comit-network/comit-rs/issues/672)
+
+## Context
+
+Currently, our node's 32 byte secret seed is stored in the config file as hex and we derive child seeds from it for each protocol execution.
+This system was invented ad-hoc and the motivation for this spike is to design a well thought out solution that is informed by other projects.
+
+I present the conclusions first and a summary of what I learned from LND.
+
+## Secret Seed Storage
+
+This section recommends how we should generate and store the secret seed by answering key questions.
+
+### Is there any way to recover anything if you have your secret seed but no database?
+
+**In practice, No.**
+
+In Theory and in the case of Ethereum only you could scan the blockchain for any HTLC looking contract paying money to an address you generated deterministically from your seed and reclaim it.
+
+In Bitcoin you can't because you'll never be able to compute the P2WSH scriptPubKey without knowing the other person's input.
+If it has already been redeemed, you could scan for redeem transactions with script that matches our Bitcoin HTLC, recover the secret and then look for the HTLC on the other chain (as above, assuming it's Ethereum).
+
+I don't think this partial recovery is worth putting the significant effort that it would require to implement.
+
+### Should it be encrypted?
+
+I think this should be optional.
+A COMIT node doesn't own the funds long term.
+It can only control the funds of protocols that are in being executed.
+This is different to LND where if someone gets access to your seed they can take all your funds regardless of whether you have channels open or not.
+
+The best attack vector against an unencrypted seed is for the attacker to get short term access to the system COMIT is running on, covertly steal their secret seed and then wait for them to execute a swap and then redeem to your own address.
+Encrypting the seed does give you protection against this.
+
+Having said that, giving an attacker access to your COMIT node will probably give them access to the HTTP API which they can use to redeem funds to their own address.
+This is probably a better attack vector than the above.
+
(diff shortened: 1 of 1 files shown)


0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+## Lightning/LND
+
+To help inform the above decisions I looked at LND because it's the flagship layer two protocol node and has to generate many keys and secrets.
+It faces a similar problem to us in that you can't restore stuff you've learned from the other party (signatures on states, revocation keys etc) from the seed.
+
+### Seed srorage
+
+The wallet seed is stored encrypted with a password in the node's `wallet.db` (or at least I think that's where it's hiding).
+You have to unlock the wallet on lnd startup with the password via an RPC call (this is what the ln-cli does).
+There is a `--noseedbackup` option which id going to be depreciated which encrypts it with the default password so you don't have unlock the wallet.
```
thomaseizinger:
```suggestion
There is a `--noseedbackup` option which is going to be depreciated which encrypts it with the default password so you don't have unlock the wallet.
```

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+[Length extension](https://en.wikipedia.org/wiki/Length_extension_attack) attacks break (2) and as we are using SHA-256 we could be susceptible to them.
+A contrived example is that let's say you derive `secret` and `secret_key` by doing `hash(seed || "secret")` and `hash(seed || "secret_key")`.
+With SHA256 it is possible to figure out what `secret_key` is once you've seen `secret` IF the length of `seed || secret` happens to be equal to the block length of the hash function (I think).
+
+Obviously, in our case you can't do this attack but I'd suggest we use a hash function that isn't susceptible to it.
+
+### Use u8 rather than strings to identify key types
+
+There's nothing inherently wrong with using utf-8 encoded strings to a hash function to produce secret keys etc. e.g. `hash(seed || "secret")` (putting length extension attacks to the side).
+This is what we do now in rfc003's [SecretSource](https://github.com/comit-network/comit-rs/blob/master/application/comit_node/src/swap_protocols/rfc003/secret_source.rs).
```
thomaseizinger: :x: It would be good to use an absolute URL here (with a commit-hash instead of master) so that the link stays valid as we evolve the code.

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+``` rust
+enum SecretType {
+    Secret,
+    RedeemKey,
+    RefundKey,
+}
+
+..
+fn secp256k1_redeem(&self) -> KeyPair {
+    KeyPair::from_secret_key_slice(self.sha256_with_seed(&[SecretType::RedeemKey as u8]).as_ref())
```
thomaseizinger, line 112: :x: 

`as u8` depends on the order though I guess? We should probably just assign constant values to it to make it totally refactoring safe.

thomaseizinger (changes requested):
Really good work!

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+Currently, our node's 32 byte secret seed is stored in the config file as hex and we derive child seeds from it for each protocol execution.
+This system was invented ad-hoc and the motivation for this spike is to design a well thought out solution that is informed by other projects.
+
+I present the conclusions first and a summary of what I learned from LND.
+
+## Secret Seed Storage
+
+This section recommends how we should generate and store the secret seed by answering key questions.
+
+### Is there any way to recover anything if you have your secret seed but no database?
```
bonomat:
```suggestion
### Is there a way to recover anything if you have your secret seed but no database?
```

bonomat:
I'm about to start reviewing this PR, but in the meantime: have you seen this?
https://github.com/lightningnetwork/lnd/blob/master/docs/recovery.md

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+
+In Bitcoin you can't because you'll never be able to compute the P2WSH scriptPubKey without knowing the other person's input.
+If it has already been redeemed, you could scan for redeem transactions with script that matches our Bitcoin HTLC, recover the secret and then look for the HTLC on the other chain (as above, assuming it's Ethereum).
+
+I don't think this partial recovery is worth putting the significant effort that it would require to implement.
+
+### Should it be encrypted?
+
+I think this should be optional.
+A COMIT node doesn't own the funds long term.
```
bonomat:
```suggestion
A COMIT node doesn't own any funds long term.
```

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+It can only control the funds of protocols that are in being executed.
+This is different to LND where if someone gets access to your seed they can take all your funds regardless of whether you have channels open or not.
+
+The best attack vector against an unencrypted seed is for the attacker to get short term access to the system COMIT is running on, covertly steal their secret seed and then wait for them to execute a swap and then redeem to your own address.
+Encrypting the seed does give you protection against this.
+
+Having said that, giving an attacker access to your COMIT node will probably give them access to the HTTP API which they can use to redeem funds to their own address.
+This is probably a better attack vector than the above.
+
+note: We could prevent this attack by giving the node a way of generating the final redeem addresses without getting them from the api.
```
bonomat:
```suggestion
Note: We could prevent this attack by allowing the node to generate the final redeem addresses without getting them from the api.
```

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+
+I think we should leave it unencrypted for now, make encryption an option later on and then perhaps make encryption the default as a last step.
```
bonomat:
```suggestion
For the sake of efficiency I propose to incrementally increase the security, i.e. (i) leave the secret unencrypted for now, (ii) make encryption an option later on and (iii) turn on encryption by default eventually.
```

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Where to store it?
+
+Unlike lnd, I don't think we should store secret seed in the database file.
+This allows users to backup the secret seed separately from the database (although they need to backup both).
```
bonomat: 👍I agree :)

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+Given that the two are linked, we should store a hash of the secret seed along with the swap id (in the same row or whatever).
+This allows the user to change the secret seed while not invalidating the entire database.
+
+By default I'd put it somewhere like:
```
bonomat: I would rather not. If it is on a fixed location it's more vulnerable for attacks as the attacker would know where to look for the secret.

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+
+```
+~/.comit/secret_seed
+```
+
+But its location should be configurable with `secret_seed_file` in the `.toml` config.
```
bonomat, line 56: 👍That sounds like the best way to go.

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+
+The format of the file should be toml with a single entry `secret_seed` (so we can `encrypted_secret_seed` later).
```
bonomat: What do you think of allowing to pass in an encrypted file itself instead?

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+
+We should make the permissions restrictive so only the user has read permission on it.
+
+#### What about encrypting the database?
+
+I don't think it's worth encrypting the database yet until we have a motivation to do it.
+An attacker stealing the database would learn what protocol executions you're involved in so atm it's just a privacy issue.
+
+We *could* encrypt the database we the `secret_seed` seeing as they're separated.
```
bonomat: Can you rephrase this sentence please (not sure if I understand it)?

bonomat (commented):
I like it. Well done :)

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,195 @@
...
+Having said that, giving an attacker access to your COMIT node will probably give them access to the HTTP API which they can use to redeem funds to their own address.
+This is probably a better attack vector than the above.
+
+note: We could prevent this attack by giving the node a way of generating the final redeem addresses without getting them from the api.
+
+I think we should leave it unencrypted for now, make encryption an option later on and then perhaps make encryption the default as a last step.
+
+### Where to store it?
+
+Unlike lnd, I don't think we should store secret seed in the database file.
```
thomaseizinger: The way this sentence is phrased confused me a bit :)

```suggestion
Unlike lnd, I think we should not store secret seed in the database file.
```

0003-secret-seed-storage-and-key-derivation.md
thomaseizinger: I am also in favor of just passing in an encrypted file that contains the seed in binary format. No need for a toml IMO.

0003-secret-seed-storage-and-key-derivation.md
thomaseizinger: The unix way to go seems to be a fixed location but restrict access to the file.
I like that better.

0003-secret-seed-storage-and-key-derivation.md
thomaseizinger: I think he wants to say:

```suggestion
We *could* encrypt the database with the `secret_seed` seeing as they're separated.
```

0003-secret-seed-storage-and-key-derivation.md
LLFourn: "giving" had the meaning I wanted to convey. In order for it to do this it would have to have a way to generate it given to it from the outside.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: I'm not sure I understood the suggestions above. What about if it's not encrypted? Do you mean in the main config file you put `secret_seed_file = /path/to/file`  OR `encrypted_secret_seed_file = /path/to/encrypted_file`.

Although this would work, I think I'd prefer having it in a toml instead because then you can transition from encrypted to unencrypted without having to edit your main config file.

thomaseizinger: Instead of having two toml entries, I'd rather design the seed-file itself in a way so that you can tell wether or not it is encrypted.

One way would be file-extension. A header in the file itself would also work.
In any case, that seems to be a much smoother way for the user to transition, since they don't have to change the config at all and there is always just one entry that refers to the `seed_file`.

LLFourn: I'd prefer not to invent a new file format and just use something that exists.
Toml certainly might not be the best format!

LLFourn:
@bonomat  No I hadn't. That page has the results of some nice lessons they've learned esp about how to backup. We could take their idea about having a restoration friendly version of the database that just has the minimal information needed to restore the protocol executions.

I've incorporated feedback (sometimes by re-writing the sentences that had suggestions). I also added a note about storing a checksum along with the seed.

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+
+```
+~/.comit/secret_seed
+```
+
+But its location should be configurable with `secret_seed_file` in the `.toml` config.
+
+The format of the file should be toml with a single entry `secret_seed` (so we can `encrypted_secret_seed` later).
+
+We should make the permissions restrictive so only the user has read permission on it.
```
D4nte, line 60: I assume you mean that if the unix permission are `og+r` then `comit_node` should reject the seed and fail to start? similar to ssh?

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+
+We can make a tool to generate/restore it from a mnemonic and encrypt/decrypt it later on.
+
+### Should it have a checksum?
+
+This probably won't catch much but it seems like a decent idea to attach a checksum to it if it's stored or transferred in a way that could be corrupted (or perhaps corrupted accidentally via a text editor).
+
+## Secret/NodeID Derivation
+
+I think the way we are doing it now is roughly the right way but I recommend modifying it in the following ways:
```
D4nte, line 81: I expect the `NodeID` to be the `libp2p id`. What relation/synergy do you propose between both? How is the `libp2p id` set?

0003-secret-seed-storage-and-key-derivation.md
D4nte: I don't think a fix location is an issue.

0003-secret-seed-storage-and-key-derivation.md
D4nte: I am happy to leave it to the user (ie, parano user choose a custom location, others choose the default one).

0003-secret-seed-storage-and-key-derivation.md
D4nte: Can someone explain what cons they see for a Toml file?

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+Currently, our node's 32 byte secret seed is stored in the config file as hex and we derive child seeds from it for each protocol execution.
+This system was invented ad-hoc and the motivation for this spike is to design a well thought out solution that is informed by other projects.
+
+I present the conclusions first and a summary of what I learned from LND.
+
+## Secret Seed Storage
+
+This section recommends how we should generate and store the secret seed by answering key questions.
+
+### Can you recover anything if you have your secret seed but no database backup?
```
bonomat, line 19: Anything in this case is too generic, I firstly wanted to ask you to explain what anything is, but then it became clear when I continued reading, hence, I think we can just remove it :)
```suggestion
### Can you recover if you have your secret seed but no database backup?
```

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+
+I don't think we should store secret seed in the database file (this is what lnd does).
+This lets users backup the secret seed separately from the database (although they need to backup both).
+Given that the two are linked, we should store a hash of the secret seed along with the swap id (in the same row or whatever).
+This allows the user to change the secret seed while not invalidating the entire database.
+
+By default I'd put it somewhere like:
+
+```
+~/.comit/secret_seed
```
bonomat: Our current configs are in: 
```
 ~/.config/comit_node/default.toml
```
should we put the `secret_seed` file in there as well?

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+- It's been well studied.
+- It's immune to length extension attacks.
+- It can produce hashes of variable length without truncating the output.
+- It's fast. Since we don't store the results of what we hash but rather derive it on demand this is handy.
+
+BLAKE2b's key is 64-bytes so I also suggest changing the secret seed we use to 64-bytes.
+
+The implementation I recommend using is https://github.com/RustCrypto/hashes just because it's maintained by the rust crypto organisation.
+
+### NodeID
```
bonomat, line 140: See comment from @D4nte above, I thought we already accepted the fact that the node ID is derived by libp2p?

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+## Lightning/LND
+
+To help inform the above decisions I looked at LND because it's the flagship layer two protocol node and has to generate many keys and secrets.
+It faces a similar problem to us in that you can't restore stuff you've learned from the other party (signatures on states, revocation keys etc) from the seed.
+
+### Seed storage
+
+The wallet seed is stored encrypted with a password in the node's `wallet.db` (or at least I think that's where it's hiding).
+You have to unlock the wallet on lnd startup with the password via an RPC call (this is what the ln-cli does).
+There is a (to be depreciated) `--noseedbackup` option  which encrypts it with the default password so you don't have unlock the wallet.
```
bonomat:
```suggestion
There is a (to be deprecated) `--noseedbackup` option  which encrypts it with the default password so you don't have unlock the wallet.
```

D4nte (changes requested):
See previous comments.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: I'd say by default the `secret_seed` can go in the same place as the config file.

Hmm is `~/.config` a thing? It seems like it's more conventional to put things in `~/.comit`.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: Interesting idea. I wasn't thinking that actually. Just that when it's first created it should be created with restrictive permissions. It is much less clear in the case of comit what the correct permissions are for the file and I would just let the user decide but provide them with sane defaults.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: The node ID is derived from the key pair derived from the seed.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: see answer to @bonomat below.

0003-secret-seed-storage-and-key-derivation.md
D4nte: looks like `~/.config` is a JS thing as I have `yarn` and Electron stuff in it.
+1 for `~/.comit` across the board.

0003-secret-seed-storage-and-key-derivation.md
D4nte: Are you saying there are **no** relations between the `NodeId` and the `libp2p id`?

0003-secret-seed-storage-and-key-derivation.md
LLFourn: I am asking: Is there anything you can recover from the secret seed by itself. 
Without it,  it implies that I am asking: Can you make a full recovery from the secret seed by itself.

0003-secret-seed-storage-and-key-derivation.md
thomaseizinger: `~/.config` is the default XDG_CONFIG_HOME directory and I guess the standard for most major linux distributions. https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

thomaseizinger: Most projects do whatever they want usually, hence home-folders are cluttered with `.my_random_project` folders. I am inclined to follow platform-friendly defaults and use something like https://github.com/soc/directories-rs to make it easily cross-platform.

On the other hand, that seems like a lost fight anyway, so if y'all vote for `.comit`, I am ok with that. Just wanted to raise some awareness :)

0003-secret-seed-storage-and-key-derivation.md
thomaseizinger: No. Libp2p creates an ID by deriving something from a keypair. We provide the keypair by deriving it from our seed.

0003-secret-seed-storage-and-key-derivation.md
thomaseizinger: > I'd prefer not to invent a new file format and just use something that exists.

I am pretty sure there are standard we can use for this.

> Can someone explain what cons they see for a Toml file?

Toml is a format for storing structured data. If all we have is a single entry, such a file format seems ill-fitted and overly complicated.

0003-secret-seed-storage-and-key-derivation.md
D4nte: https://github.com/soc/directories-rs looks great

0003-secret-seed-storage-and-key-derivation.md
D4nte: I think I don't understand what is the point of this new `NodeID`

thomaseizinger: The ID used by libp2p IS the node id!

D4nte: > The ID used by libp2p IS the node id!

I expect this to be stated in the document and not something that takes 24 hours to understand/decrypt.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: I like the idea of using directories-rs, but putting crap in `/User/Name/Library` on mac is kind offensive. It gives us an answer for what to do on windows though which is nice.

D4nte: I don't mind crap in `~/Library/` on Mac, isn't it the standard?

0003-secret-seed-storage-and-key-derivation.md
LLFourn: I added the sentence: `This is the NodeID that will be used by libp2p.`.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: @thomaseizinger The thing is we have at least two entries: `secret_seed` and `encrypted_secret_seed`. We could use [pem](https://docs.rs/pem/0.6.0/pem/) to label one or the other.

0003-secret-seed-storage-and-key-derivation.md
bonomat: > On the other hand, that seems like a lost fight anyway, so if y'all vote for .comit, I am ok with that. Just wanted to raise some awareness :)

Don't give up that fast, you have me on your side: 
**+1 on `~/.config/comit_node`**

D4nte: Let's aim for a `Win/Win` deal :)

LLFourn: I removed mention of the path of secret_seed file and just said it should be next to the config file by default. Discussion can happen elsewhere about where to actually put the config file.

0003-secret-seed-storage-and-key-derivation.md
LLFourn: I changed it to say the format should be text like the pem format.

bonomat (approved):

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+
+The best attack vector against an unencrypted seed is for the attacker to get short term access to the system COMIT is running on, covertly steal their secret seed and then wait for them to execute a swap and then redeem to their own address.
+Encrypting the seed does give you protection against this.
+
+Having said that, giving an attacker access to your COMIT node will probably give them access to the HTTP API which they can use to redeem funds to their own address.
+This is probably a better attack vector than the above.
+
+Note: We could prevent this attack by giving the node a way of generating the final redeem addresses without getting them from the api.
+
+For the sake of expediency I propose we incrementally increase the security, i.e. (i) leave the secret unencrypted for now, (ii) make encryption an option later on and (iii) make encryption the default eventually.
```
bonomat, line 45: 👍

0003-secret-seed-storage-and-key-derivation.md
```diff
@@ -0,0 +1,201 @@
...
+#### What about encrypting the database?
+
+I don't think it's worth encrypting the database yet until we have a motivation to do it.
+An attacker stealing the database would learn what protocol executions you're involved in so atm it's just a privacy issue.
+
+We *could* encrypt the database with the `secret_seed` seeing as they're separated.
+
+### How to generate it?
+
+If `secret_seed_file` isn't set, I think we should just generate it from `OsRng` to `~/.comit/secret_seed` if it doesn't exist.
```
bonomat: We can rediscuss the location once we go for the implementation. Obviously, if we move to `~/.comit/`, then the config file should go there as well.

D4nte (approved):

0003-secret-seed-storage-and-key-derivation.md
LLFourn: I removed that mention of `~/.comit` replacing it with "its default location' :)

thomaseizinger (approved):

PR #6, opened 2019-04-24
Add spike for LN as alpha ledger

### LINKED ISSUE #841: ⚗️Investigate compatibility of RFC003 and LND with HODL invoices
This Spike is about figuring out how we *could* do an [RFC003](https://github.com/comit-network/RFCs/blob/master/RFC-003-SWAP-Basic.md) SWAP with `lnd` as one of the `Ledger`s. Note that this would be independent of lightning channels as a Ledger (unless you find that lnd exposes an api that would allow this).

HODL invoices were introduced into `lnd` here: https://github.com/lightningnetwork/lnd/pull/2022 which will make things easier.

What to include:
1. A sketch of the protocol (the actions involved at each ledger state, what the SWAP request response would look like). Justify your decisions.
2. A summary of architectural changes we'd need to make to realise the proposal.
3. An analysis of any inconsistencies the proposal would have with our present idea of RFC003.

bonomat:
Spike for supporting Lightning Network as Alpha Ledger.

Resolves comit-network/comit-rs#841


0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Context
+
+The **Hold invoice feature** was introduced in the following PR: [lightningnetwork/lnd/pull/2022](https://github.com/lightningnetwork/lnd/pull/2022).
+> A hold invoice is a new type of invoice that triggers a different flow at the receiver end.
+> Instead of immediately locking in and settling the htlc when the payment arrives, the htlc for a hold invoice is only locked in and not yet settled.
+> At that point, it is not possible anymore for the sender to revoke the payment, but the receiver still can choose whether to settle or cancel the htlc and invoice.
+
+This allows us to trade Bitcoin on the Lightning Network (LN) for an asset _X_ on a different ledger where LN is the _alpha ledger_ as per the definition in [RFC003](https://github.com/comit-network/RFCs/blob/master/RFC-003-SWAP-Basic.md).
+For the other way around, Bitcoin on the Lightning Network is the _beta asset/ledger_, we can make use of `createinvoice`.
```
D4nte:
```suggestion
For the other way around, Bitcoin on the Lightning Network is the _beta asset/ledger_, we can make use of `createinvoice` as Alice would instantly release the secret by settling Bob's LN payment
```

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Standard _hold invoice_ process
+
+> Assumption: Alice wants to pay Bob _X_ Bitcoin using the LN but Bob does not generate the preimage - Alice does it.
+1) Alice generates a secret preimage `p` and sends the hash of it `h(p)` to Bob.
+2) Bob calls `addholdinvoice` with an amount `X` and the hashed preimage `h(p)` ([more details](https://github.com/lightningnetwork/lnd/blob/aa1cd04dbf07a9195d5ada752f383988d8d01fa7/cmd/lncli/invoicesrpc_active.go#L142)).
+3) Bob sends the generated invoice to Alice.
+4) Alice now can pay the invoice using `sendpayment`.
+5) As soon as Alice is happy (because she is in a good mode), she tells Bob the secret (or he learns it otherwise, see below).
```
D4nte:
```suggestion
5) As soon as Alice is happy (because she is in a good mood), she tells Bob the secret (or he learns it otherwise, see below).
```

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+6) Bob settles the payment by calling `settleInvoice` ([more details](https://github.com/lightningnetwork/lnd/blob/aa1cd04dbf07a9195d5ada752f383988d8d01fa7/cmd/lncli/invoicesrpc_active.go#L53)).
```
D4nte: I would not consider the spike complete until this is tested and we know how it looks like (at least on command line).

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+
+The payment is complete.
+
+## Research
+
+> Assumption: Bob wants to receive _X_ Bitcoin (**A** Asset) on the LN (**alpha** Ledger) for _Y_ Ether (**B** Asset) on Ethereum (**beta** Ledger)
+> Alice magically finds out about this _whish_ or using a [COMIT link](fill in some link).
```
D4nte:
```suggestion
> Alice magically finds out about this _wish_ or using a [COMIT link](fill in some link).
```

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+>
+> * Alpha Ledger: Lightning Network
+> * Alpha Asset: Bitcoin on the Lighting Network
```
D4nte:
```suggestion
> * Alpha Asset: Bitcoin
```

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+      * `SubscribeSingleInvoice` - this is not available through the CLI but as RPC; or
+      * `LookupInvoice` - this is available trough the CLI but needs to be polled regularly.
+3) Bob response to Alice's request with the following information:
+
+| Name                    | JSON Encoding | Description                                                                                               |
+|-------------------------|---------------|-----------------------------------------------------------------------------------------------------------|
+| `alpha_redeem_identity` | `α::Identity` | The identity on alpha that **A**, this is the LN `invoice`  |
+| `beta_refund_identity`  | `β::Identity` | The identity on beta that **B** will be transferred to when the beta-HTLC is activated after `beta_expiry`      |
+
+4) Alice now starts the [exection phase](https://github.com/comit-network/RFCs/blob/master/RFC-003-SWAP-Basic.md#1-alice-deploys-%CE%B1-htlc) by paying the invoice using the LND command `sendpayment`.
```
D4nte:
```suggestion
4) Alice now starts the [execution phase](https://github.com/comit-network/RFCs/blob/master/RFC-003-SWAP-Basic.md#1-alice-deploys-%CE%B1-htlc) by paying the invoice using the LND command `sendpayment`.
```

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Considerations and limitations
+
+#### Lightning Network: a new ledger
+Similiar to the ledger definitions for [Bitcoin](https://github.com/comit-network/RFCs/blob/master/RFC-004-Bitcoin.md) and [Ethereum](https://github.com/comit-network/RFCs/blob/master/RFC-006-Ethereum.md) we need to handle the Lightning Network differently.
+This is required because the comit-node and btsieve need to perform different actions accordingly.
+We are always talking about Ledgers and Assets, (e.g. _Bitcoin_ Asset on the _Bitcoin_ Ledger, _Ether_ Asset on the _Ethereum_ Ledger, _Erc20_ on the _Ethereum_ Ledger, ...), Hence,
+if we follow this approach, for supporting LN (through LND) we will need to introduce a new pair of Ledger and Asset:
+
+* Ledger: the **Lightning Network**. _Ledgers_ are used as _settlement layers_ for our HTLCs. In the case of LND, this layer is the Lightning Network.
```
D4nte: One could argue whether it should be `LightningNetwork` or `LND`. I opt for `LightningNetwork` at this stage.

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+* Asset: **Bitcoin**. Since LN is a layer-2 network on top of Bitcoin, the asset should also be Bitcoin.
+
+#### Timeouts
+When calling `addholdinvoice` an [`expiry`](https://github.com/lightningnetwork/lnd/blob/aa1cd04dbf07a9195d5ada752f383988d8d01fa7/cmd/lncli/invoicesrpc_active.go#L142) (in seconds) can be given. However, this timeout is relative and will get accumulated across the payment path.
```
D4nte: This `expiry` has no blockchain consequences. It is just to indicate to the payer that the `invoice` is expired (it is not an HTLC expiry) and that they should request a new invoice from the payee.

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+Hence, Alice will need to know the path and this value in advance before sending a swap request to Bob as she won't (or should not) be able to change neither `alpha_expiry` nor `beta_expiry` after the swap request.
```
D4nte: This is a combination of parameters that are configured globally for the payee node and I think on-the-go (when paying) for the payer node. Need to dig in my notes.

The payer configures with `--final_cltv_delta=T` when paying and the payee configures when opening a channel or global conf for the lnd.

It means that the payee needs to accept/propose an expiry that matches their lnd conf.

It means that the payer could deviate from the agreed expiry and lnd may:
- accept the invoice if it is still above the configure threshold (CSV I think)
- reject the invoice automatically if it is under the threshold.

but then, the HTLC details are visible to the payee so the payee could decide not to proceed with the swap if the expiry does not match expectations

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+
+#### Responsabilitites
```
D4nte:
```suggestion
#### Responsibilities
```

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,158 @@
...
+
+A main goal of COMIT is to keep the autonomy to the user and let him/her decide when to deploy a HTLC, redeem or refund a HTLC, etc.
+If a trade involves LN using LND we can aproach these things differently:
```
D4nte:
```suggestion
If a trade involves LN using LND we can approach these things differently:
```

0003-lightning-as-alpha-ledger.md
bonomat: Fair enough, I already started playing around with it :)

0003-lightning-as-alpha-ledger.md
bonomat: I also opt for `LightningNetwork`.

0003-lightning-as-alpha-ledger.md
bonomat: Thanks for pointing this out.

0003-lightning-as-alpha-ledger.md
bonomat: Thanks for pointing this out.

> It means that the payee needs to accept/propose an expiry that matches their lnd conf.

This would not fit into our model right? Bob is the payee and Alice the one who proposes the timeouts in her swap request.

D4nte: In the case of LN being alpha ledger then it's fine: It means that the payee needs to accept ~/propose~ an expiry that matches their lnd conf.

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+* Action
+    * Create hold invoice
+* Responsibility
+    * LND
+* Invoker
+    * User, comit-i, comit-node
+* Description
+    * `addholdinvoice` is available as a RPC command or through the LND CLI. Althoug dealing with this is rather cumbersome, to keep the autonomy with the user, and to not introduce LND dependency into the comit-node, we this should be possible through comit-i.
```
D4nte:
```suggestion
    * `addholdinvoice` is available as a RPC command or through the LND CLI. Although dealing with this is rather cumbersome, to keep the autonomy with the user, and to not introduce LND dependency into the comit-node, we this should be possible through comit-i.
```

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,164 @@
...
+* Conclusion:
+    * comit-i needs LND support
```
D4nte: Ok but do we want comit-i to do it automatically? What are the implications on comit-rs HTTP API?

Should have some sequence graph to help us see any action dichotomy?

0003-lightning-as-alpha-ledger.md
```diff
@@ -0,0 +1,164 @@
...
+* Action
+    * Monitor LN
+* Responsibility
+    * LND
+* Invoker
+    * Btsieve
+* Description
+    * Similar to other Ledgers we need to monitor LN for the payment (and later on settlement) of an invoice. To keep our current abstraction layer, this should be done through btsieve
+* Conclusion
+    * btsieve needs LND support
```
D4nte: I am not sure about this one as it does not really fit the btsieve model so far and I thought that is where we left the discussion.

Especially that currently btsieve could be publicly deployed but then, for LN, it must be a private instance.

However, thanks to comit-network/comit-rs#874 it may be ok to integrate LN as btsieve. Someone could run a public btsieve for Bitcoin and Ethereum and users would run a private btsieve for LN.

This would mean that it would be interesting to configure comit-rs to have a global configuration for btsieve endpoint but also a per ledger one so that comit-rs can connect to these two btsieves.

0003-lightning-as-alpha-ledger.md
bonomat: I can confirm, the following works: 


`secret=0000000000000000000000000000000000000000000000000000000000000001`

`sha256=ec4916dd28fc4c10d78e287ca5d9cc51ee1ae73cbfde08c6b37324cbfaac8bc5`

---

```bash
$bob: lncli --network=simnet addholdinvoice ec4916dd28fc4c10d78e287ca5d9cc51ee1ae73cbfde08c6b37324cbfaac8bc5 --amt 10
```



---



```bash
$alice: lncli --network=simnet payinvoice lnsb100n....
```


---

```bash
$bob: `lncli --network=simnet lookupinvoice 4146873...
{
    "memo": "",
    "receipt": null,
    "r_preimage": null,
    ...
    "amt_paid_msat": "10000",
    "state": "ACCEPTED"
}
```

---

```bash
$bob: lncli --network=simnet settleinvoice 0000000000000000000000000000000000000000000000000000000000000001
``` 

---

```bash
$bob: `lncli --network=simnet lookupinvoice ec4916dd28fc4c10d78e287ca5d9cc51ee1ae73cbfde08c6b37324cbfaac8bc5
{
    "memo": "",
    "receipt": null,
    "r_preimage": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAE=",
    "r_hash": "7EkW3Sj8TBDXjih8pdnMUe4a5zy/3gjGs3Mky/qsi8U=",
    "value": "10",
    "settled": true,
    "creation_date": "1556272625",
    "settle_date": "1556272661",
    "payment_request": "lnsb100n1pwv9403pp5a3y3dhfgl3xpp4uw9p72tkwv28hp4eeuhl0q334nwvjvh74v30zsdqqcqzpgjemg8uy8y7ej0q6lmxvkfhwfrjnesv6jryk46m5kcfcd2drykskq97c2ad9lvp9d6mmyt0r6rhp26e2cmrqd9qgc88rf5l58hz6ntxgqc3azkn",
    "description_hash": null,
    "expiry": "3600",
    "fallback_addr": "",
    "cltv_expiry": "40",
    "route_hints": [
    ],
    "private": false,
    "add_index": "7",
    "settle_index": "1",
    "amt_paid": "10000",
    "amt_paid_sat": "10",
    "amt_paid_msat": "10000",
    "state": "SETTLED"
}
```

0003-lightning-as-alpha-ledger.md
bonomat: No, not automatically. What I meant was that comit-i should support LN in a similar manner as it supports Ethereum (e.g. through a browser extension like metamask) now.

0003-lightning-as-alpha-ledger.md
bonomat: > Especially that currently btsieve could be publicly deployed but then, for LN, it must be a private instance.

That's true, but I think it fits better into our abstraction layer.

0003-lightning-as-alpha-ledger.md
D4nte: should we do a summary of the steps needed to execute rfc003?

bonomat: checkout the description above "Step-by-step description". Should be rfc003 steps.

D4nte (approved):
Great and exciting stuff!

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,277 @@
...
+
+* Alpha Ledger: Lightning Network
+* Alpha Asset: Bitcoin
+* Beta Ledger: Ethereum
+* Beta Asset: Ether
+____
+
+=== Step-by-step description
+
+. Alice generates a secret preimage `p` and sends a swap request to Bob, this includes (assets with amounts are emitted):
```
bonomat:
```suggestion
1. Alice generates a secret preimage `p` and sends a swap request to Bob, this includes (assets with amounts are emitted):
```

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,277 @@
+= LND as Alpha Ledger as described in RFC003
+
+NOTE: Authors: Philipp Hoenisch +
+Date: 2019-04-24 +
+Tracking issue: TBA
+
```
bonomat, line 8:
```suggestion
:toc:
```

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,279 @@
...
+=== Standard _hold invoice_ process
+
+____
+Assumption: Alice wants to pay Bob _X_ Bitcoin using the LN but Bob does not generate the preimage - Alice does it.
+____
+
+1) Alice generates a secret preimage `p` and sends the hash of it `h(p)` to Bob.
+2) Bob calls `addholdinvoice` with an amount `X` and the hashed preimage `h(p)` (https://github.com/lightningnetwork/lnd/blob/aa1cd04dbf07a9195d5ada752f383988d8d01fa7/cmd/lncli/invoicesrpc_active.go#L142[more details]).
+3) Bob sends the generated invoice to Alice.
```
thomaseizinger, line 35: Is this necessary? Alice should be able to generate that invoice by herself right?

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+[cols=3*,options=header]
+|===
+| Name
+| JSON Encoding
+| Description
+
+| `alpha_redeem_identity`
+| `α::Identity`
+| The identity on alpha that **A**, this is the LN `invoice`
```
thomaseizinger: This seems very hacky. An invoice is not an identity. This should be Bob's Lightning node ID so that Alice knows, where to send the money to. She can then construct the invoice herself if that is what LND needs to make a payment.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+== Spike Outcome
+
+=== Lightning Network: a new ledger
+Similiar to the ledger definitions for https://github.com/comit-network/RFCs/blob/master/RFC-004-Bitcoin.md[Bitcoin] and https://github.com/comit-network/RFCs/blob/master/RFC-006-Ethereum.md[Ethereum] we need to handle the Lightning Network differently.
+This is required because the comit-node and btsieve need to perform different actions accordingly.
+We are always talking about Ledgers and Assets, (e.g. _Bitcoin_ Asset on the _Bitcoin_ Ledger, _Ether_ Asset on the _Ethereum_ Ledger, _Erc20_ on the _Ethereum_ Ledger, ...), Hence,
+if we follow this approach, for supporting LN (through LND) we will need to introduce a new pair of Ledger and Asset:
+
+* Ledger: the **Lightning Network**. _Ledgers_ are used as _settlement layers_ for our HTLCs. In the case of LND, this layer is the Lightning Network.
```
thomaseizinger, line 209: :x: I think this spike should include, how the ledger is exactly represented.

With the LN, we have node IDs and channels. A channel is a construct between two nodes.
It looks like the `Ledger` definition cannot be a node ID, it has to be the channel id, otherwise which node ID do you use? Alice's or Bob's?
If we refer to the channel, Alice aswell as Bob should be able to query their local node to see if they are in the specified channel and can also find out the other party's node ID (at least I hope LND exposes this information :upside_down_face:)

This in turn makes the alpha redeem and success identity redundant, since both would just either the channel or node ID which are already known information at this point.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,279 @@
...
+* Action
+    ** Create hold invoice
+* Responsibility
+    ** LND
+* Invoker
+    ** User, comit-i, comit-node
+* Description
+    ** `addholdinvoice` is available as a RPC command or through the LND CLI. Although dealing with this is rather cumbersome, to keep the autonomy with the user, and to not introduce LND dependency into the comit-node, we this should be possible through comit-i.
+* Conclusion:
+    ** comit-i needs LND support. However, in order to do this, we will need to introduce a new action which is meant to be executed prior accepting a swap request:
```
thomaseizinger, line 232: Meh, I disagree. If things have to be done atomically (sending accept + adding the invoice) then you should really build it in a way so that the user does _NOT_ have a choice to break this invariant. For me, such atomicity is not guaranteed just because comit-i exposes it as a single action. From that, I would conclude that the comit-node should perform the `addholdinvoice` action by talking to LND. Thinking ahead (comit-node speaking L1 and possibly L2 protocols directly), this seems to be a path that makes such a transition much easier.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+
+* Action:
+    ** Settle Invoice
+* Responsibility
+    ** LND or LN Wallet
+* Invoker
+    ** User, comit-i, comit-node
+* Description
+    ** As soon as the secret has been learned, the HTLC on the LN should be settled using the command `settleinvoice`, this can either be done by the user (and exposed through comit-i) or done automatically through the comit-node. Since we have the extra _redeem_ step for Bitcoin and Ethereum (as well for Erc20) which needs to be performed from the user, we should leave the settlement of the invoice to the user (e.g. expose this information through comit-i).
```
thomaseizinger: An important thing to note here is that one of the reasons we have an extra redeem step is because we cannot send transactions from the comit-node (we don't have funds). For LN, this restriction is not given since we don't have to pay any fees to settle an invoice.

I am not suggesting that we should change our behaviour because of that. It is just something to keep in mind. Once we have L1 support, it could be worth considering to give the comit-node the functionality to own little funds, enough for sending redeem transactions. For L2 ledgers, we could then also automatically redeem funds.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+
+* Action
+    ** Monitor LN
+* Responsibility
+    ** LND
+* Invoker
+    ** Btsieve
+* Description
+    ** Similar to other Ledgers we need to monitor LN for the payment (and later on settlement) of an invoice. To keep our current abstraction layer, this should be done through btsieve
```
thomaseizinger, line 278: I disagree. For me `btsieve` is about monitoring L1. L2 doesn't have blocks so none of the abstractions built into `btsieve` fit for L2.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,279 @@
...
+* Conclusion
+    ** btsieve needs LND support.
+
+
+=== COMIT link relation
+Assumption: Bob is the creator of the link and is willing to receive Bitcoin for Ether for 1:10.
+
+Bob cannot yet create a hold invoice, i.e. `addholdinvoice` as he does not know the secret yet, hence, all he can add in the link is the information about Ledgers, Assets and exchange rates.
```
thomaseizinger: Is that different from what we are doing now? Even for none-LN swaps, the link does not contain the secret hash.

thomaseizinger (changes requested):

0003-lightning-as-alpha-ledger.adoc
D4nte: Yes @thomaseizinger is correct.

0003-lightning-as-alpha-ledger.adoc
D4nte: Yes I agree

0003-lightning-as-alpha-ledger.adoc
D4nte: I see no added value in considering channels ids in the protocol and implementation.
I do not see why Alice and Bob should query their node for a specific channel.
Started to play channel ids just add unjustified restrictions on what route should a payment come from.

I agree, identity should be node id.

Bolt Lightning over Bitcoin Core is the ledger.

0003-lightning-as-alpha-ledger.adoc
D4nte: I think it's a valid debate we discussed yesterday. Do you expose the invoice creation as an action in the HTTP API or do you have `comit_node` do it for you.

But then, if `comit_node` is the one creating the hold invoice, why not being the one doing the payment?

And if you have `comit_node` connected to lnd, then why not do the watching directly from lnd instead of having btsieve connected to local lnd?

I think the architectural point of what service talks to which service may deserve a live discussion.

0003-lightning-as-alpha-ledger.adoc
D4nte: Another point is that not doing the automated settlement allows Alice to use her option whereas if we automate it, Alice would have to use a different implementation to use her option (or fork comit-rs)

0003-lightning-as-alpha-ledger.adoc
D4nte: > Is that different from what we are doing now? Even for none-LN swaps, the link does not contain the secret hash.

Where are you reading that hte link should contain the secret hash?

D4nte:
@thomaseizinger you have quite a lot of comments that I don't think were in the original spike you wrote with @LLFourn. Is there any notes or discussions that you had that were not reported in the original spike?

0003-lightning-as-alpha-ledger.adoc
bonomat: True that, it's possible. Two points come to my mind: 
 * By doing this we identify an invoice by the `secret_hash` and not by the invoice itself ala [bolt 11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md). Meaning, Bob will need to watch for an invoice by the `secret_hash` itself -> which is fine. 
* We cannot make use of features such as a signed invoice, optional proposed routing table, description, etc, -> which might be useful in the future

0003-lightning-as-alpha-ledger.adoc
bonomat: Partially agree :P 
It's true that the naming (`*_*_identity`) does not fit perfectly as it's not _just_ an identity. 
However, the invoice contains Bob's LN node ID. 
The question we should ask ourselves is whether we can return something differently here as well instead of the pair `alpha_redeem_identity` and `beta_refund_identity` because strictly said: even Bob's LN node ID is not a `redeem_identity`.

0003-lightning-as-alpha-ledger.adoc
bonomat: > It looks like the Ledger definition cannot be a node ID, it has to be the channel id, otherwise which node ID do you use? Alice's or Bob's?

What do you mean with _which node ID do you use_ or _`Ledger` definition cannot be a node ID, it has to be the channel id_? 

> Bolt Lightning over Bitcoin Core is the ledger.

I agree on this. The channel does not seem useful to me, i.e. Alice has no guarantees to see all channels of Bob nor can she/or should she send a payment to a particular channel. Payments on LN are identified by the node IDs. 

Think differently, if Bob provides Alice a channel ID - a channel between Bob and Claire. What use is that for Alice? All she can do is verify that this channel exist, but nothing about its state, nor can she send a payment to that channel.

However, if we limit ourselves to 0-hop channels (Alice <-> Bob) than the channel might be useful, but that's not what we want.

0003-lightning-as-alpha-ledger.adoc
bonomat: I disagree with you. Btsieve knows about blocks and transactions, transactions can either be done on L2 or L1.

0003-lightning-as-alpha-ledger.adoc
bonomat: I think his comment was updated afterwards... 
You are right, this part of the spike can be removed.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,274 @@
...
+
+The **Hold invoice feature** was introduced in the following PR: https://github.com/lightningnetwork/lnd/pull/2022[lightningnetwork/lnd/pull/2022].
+____
+A hold invoice is a new type of invoice that triggers a different flow at the receiver end.
+Instead of immediately locking in and settling the htlc when the payment arrives, the htlc for a hold invoice is only locked in and not yet settled.
+At that point, it is not possible anymore for the sender to revoke the payment, but the receiver still can choose whether to settle or cancel the htlc and invoice.
+____
+
+This allows us to trade Bitcoin on the Lightning Network (LN) for an asset _X_ on a different ledger where LN is the _alpha ledger_ as per the definition in https://github.com/comit-network/RFCs/blob/master/RFC-003-SWAP-Basic.md[RFC003].
+For the other way around, Bitcoin on the Lightning Network is the _beta asset/ledger_, we can make use of `createinvoice` as Alice would instantly release the secret by settling Bob's LN payment.
```
LLFourn: Wouldn't it be better to also use hold invoices when it's the `beta_ledger` too to keep everything consistent?

0003-lightning-as-alpha-ledger.adoc
LLFourn: The idea of just passing through the invoice at the comit layer is interesting.

PS what's a signed invoice?

0003-lightning-as-alpha-ledger.adoc
LLFourn: This ambiguity really makes it seem like LN doesn't fit rfc003 🤔.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,274 @@
...
+This is required because the comit-node and btsieve need to perform different actions accordingly.
+We are always talking about Ledgers and Assets, (e.g. _Bitcoin_ Asset on the _Bitcoin_ Ledger, _Ether_ Asset on the _Ethereum_ Ledger, _Erc20_ on the _Ethereum_ Ledger, ...), Hence,
+if we follow this approach, for supporting LN (through LND) we will need to introduce a new pair of Ledger and Asset:
+
+* Ledger: the **Lightning Network**. _Ledgers_ are used as _settlement layers_ for our HTLCs. In the case of LND, this layer is the Lightning Network.
+* Asset: **Bitcoin**. Since LN is a layer-2 network on top of Bitcoin, the asset should also be Bitcoin.
+
+=== Dealing with timeouts
+When calling `addholdinvoice` an https://github.com/lightningnetwork/lnd/blob/aa1cd04dbf07a9195d5ada752f383988d8d01fa7/cmd/lncli/invoicesrpc_active.go#L142[`expiry`] (in seconds) can be given. However, this expiry is only related to the invoice itself and has nothing to do with the expiry times on the HTLCs later on.
+However, timeouts on the HTLCs are relative and will get accumulated across the payment path, which makes it harder to predict.
```
LLFourn: 🔧  Timeouts on HTLCs are absolute timeouts. They may increase relative to each other along each hop but they are an absolute block height (i.e. CLTV). I think that's what you meant but this confused me so I suggest rewording it.

0003-lightning-as-alpha-ledger.adoc
thomaseizinger: > Where are you reading that hte link should contain the secret hash?

`he does not know the secret yet, hence, all he can add in the link is the information about Ledgers, Assets and exchange rates.` <- this suggests that the secret/secret_hash should go in the link.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,274 @@
...
+Hence, Alice will need to know the path in advance before sending a swap request to Bob as she won't (or should not) be able to change neither `alpha_expiry` nor `beta_expiry` after the swap request.
+
+Timeouts on the HTLCs are configured globally for the payee node and can optionally be defined when paying by the payer node (`--final_cltv_delta=T`).
```
LLFourn: In response to the two sentences above:

I don't see why Alice needs to know the path in advance of sending a swap request. The `final_cltv_delta` thing should be enough to give a HTLC to Bob with the desired timeout (in absolute bocks instead of absolute seconds).

The caveat here is that you can't set the actual `cltv_expiry` on the last hop but only the delta. The LN node calculates what the `cltv_expiry` will be based on `final_cltv_delta` and what it perceives to be the current block height.

thomaseizinger:
> @thomaseizinger you have quite a lot of comments that I don't think were in the original spike you wrote with @LLFourn. Is there any notes or discussions that you had that were not reported in the original spike?

We never wrote a spike, we only started looking into the topic for ~1h. There were quite a few unanswered questions and open points at that point. All we wrote down was the current state, which was mostly about getting the new feature to actually work + what you can do with it.
I am bit surprised about your expectation that our spike was done/complete.

Some of the open points have now been answered/addressed in this spike, hence my comments.

0003-lightning-as-alpha-ledger.adoc
thomaseizinger: > strictly said: even Bob's LN node ID is not a redeem_identity.

Correct. It should probably not even be there (i.e. empty).

0003-lightning-as-alpha-ledger.adoc
thomaseizinger: @D4nte 

> Bolt Lightning over Bitcoin Core is the ledger.

With or without any identifiers of the nodes/channels?
We talked about this topic quite a lot already and explored different options. To me, this spike doesn't yet clarify, which of those options we should now go for and what the implications are.

thomaseizinger: @bonomat 

> Think differently, if Bob provides Alice a channel ID - a channel between Bob and Claire. What use is that for Alice? All she can do is verify that this channel exist, but nothing about its state, nor can she send a payment to that channel.
However, if we limit ourselves to 0-hop channels (Alice <-> Bob) than the channel might be useful, but that's not what we want.

True, I was only thinking 0-hop channels.

D4nte: Without any identifier to nodes or channels.

There are no identifier to Bitcoin Core Ledger except for the name `bitcoin` and the network (e.g. `mainnet`).

I expect the Lightning BOLT Bitcoin Core ledger to have the same amount of information to be identified: a name (e.g. `LightningBitcoin`) and a network (e.g. `mainnnet`).

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,274 @@
...
+    ** Btsieve
+* Description
+    ** Similar to other Ledgers we need to monitor LN for the payment (and later on settlement) of an invoice. To keep our current abstraction layer, this should be done through btsieve
+* Conclusion
+    ** btsieve needs LND support.
+
+
+=== Fall-back mechanism of LN
+LN allows to specify a fallback address (_fallback_addr_) in when creating calling `addholdinvoice`.
+We could use this information to fall back to an on-chain HTLC trade if no route can be found between Alice and Bob.
```
LLFourn, line 285: Do you have a citation for the `fall back to an on-chain HTLC trade` bit. I looked around and couldn't find that it actually uses a HTLC.

From thinking about it -- it's unlikely that it does. It's just a normal payment. This is actually extremely dangerous for us if we were to let the other party generate the invoice without checking it doesn't contain a fallback. This would mean that the LND node might be tricked into sending money on chain rather than doing a HTLC.

0003-lightning-as-alpha-ledger.adoc
LLFourn: Like Thomas, I think the idea was that `btsieve` could be run a service. This implies that it only works on public ledgers.

0003-lightning-as-alpha-ledger.adoc
thomaseizinger: Thanks! That is the information I was looking for :)

0003-lightning-as-alpha-ledger.adoc
bonomat: We can, but for now, we focus on one direction first.

0003-lightning-as-alpha-ledger.adoc
bonomat: > PS what's a signed invoice?

To verify that you are not paying an invoice of someone else?

0003-lightning-as-alpha-ledger.adoc
bonomat: You are right, it does not use HTLCs but falls back to a direct payment. However, we could use it for our HTLCs, i.e. we should not let LND handle these things for us.

D4nte:
> > @thomaseizinger you have quite a lot of comments that I don't think were in the original spike you wrote with @LLFourn. Is there any notes or discussions that you had that were not reported in the original spike?
> 
> We never wrote a spike, we only started looking into the topic for ~1h. There were quite a few unanswered questions and open points at that point. All we wrote down was the current state, which was mostly about getting the new feature to actually work + what you can do with it.
> I am bit surprised about your expectation that our spike was done/complete.
> 
> Some of the open points have now been answered/addressed in this spike, hence my comments.

Clearly a misunderstanding. I thought the take over was to organise the work to be done. I understood that the spike was finished hence my surprise of the initial content length. Now I wonder why it was handed over. Clearly I missed something🤔

thomaseizinger:
> Clearly a misunderstanding. I thought the take over was to organise the work to be done. I understood that the spike was finished hence my surprise of the initial content length. Now I wonder why it was handed over. Clearly I missed something :thinking:

IIRC it was handed over because I started to work on libp2p stuff and @LLFourn started to dig into Scriptless-Scripts.
This happened as a result of discussing, who is the most interested in working on which topics. I think you and @bonomat were keen on Lightning, hence you took over were we left of.

D4nte:
I see no pending comments, is this ready for merge?

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,354 @@
...
+1) Alice generates a secret preimage `p` and sends the hash of it `h(p)` to Bob.
+2) Bob calls `addholdinvoice` with an amount `X` and the hashed preimage `h(p)` (https://github.com/lightningnetwork/lnd/blob/aa1cd04dbf07a9195d5ada752f383988d8d01fa7/cmd/lncli/invoicesrpc_active.go#L142[more details]).
+3) Bob sends the generated invoice to Alice.
+4) Alice now can pay the invoice using `sendpayment`.
+5) As soon as Alice is happy (because she is in a good mood), she tells Bob the secret (or he learns it otherwise, see below).
+6) Bob settles the payment by calling `settleInvoice` (https://github.com/lightningnetwork/lnd/blob/aa1cd04dbf07a9195d5ada752f383988d8d01fa7/cmd/lncli/invoicesrpc_active.go#L53[more details]).
+
+The payment is complete.
+
+== Lightning Network as Alpha Ledger within the bounds of RFC003
```
thomaseizinger: :x: We settled for a different protocol. I think this should be reflected here.

0003-lightning-as-alpha-ledger.adoc
```diff
@@ -0,0 +1,354 @@
...
+Assumption: Bob wants to receive _X_ Bitcoin (**A** Asset) on the LN (**alpha** Ledger) for _Y_ Ether (**B** Asset) on Ethereum (**beta** Ledger)
+Alice magically finds out about this _wish_ or using a COMIT link.
+
+* Alpha Ledger: Lightning Network
+* Alpha Asset: Bitcoin
+* Beta Ledger: Ethereum
+* Beta Asset: Ether
+____
+
+According to https://github.com/comit-network/RFCs/blob/master/RFC-003-SWAP-Basic.md[RFC003], Alice needs to send the following information to Bob for initiating a swap.
```
thomaseizinger, line 56: Not sure if we should reference RFC003 here if we are going for a different protocol?

thomaseizinger: To avoid confusion, I think this paragraph should first mention that we _tried_ to fit LN into RFC003.

thomaseizinger (changes requested):

0003-lightning-as-alpha-ledger.adoc
bonomat: The confusion should be cleared at the end of this section:
> Because of these limitations, we should introduce a new protocol for using LN as Alpha Ledger.

do you want me to mention this already at the beginning of the section?

bonomat: ok, will do so as discussed offline.

thomaseizinger (approved):

PR #7, opened 2019-05-01
Siren `GET /swaps` response prototype

### LINKED ISSUE #917: Prototype accept/decline actions using siren
In order to move forward easily with Bobtimus prime we should define the siren json document of how a  `IN_PROGRESS` response look like in each state of accept and decline, i.e. when you call `GET /swaps/` for Bob.  

- [x] Create one document containing a swap with
  - accept action
  - decline action
- [x] validate agains siren json schema

D4nte:
Siren prototype.

Fixes comit-network/comit-rs#917


0003-siren-prototype-accept-decline.adoc
```diff
@@ -0,0 +1,88 @@
...
+        {
+          "name": "accept",
+          "title": "Accept",
+          "method": "POST",
+          "href": "/swaps/rfc003/399e8ff5-9729-479e-aad8-49b03f8fc5d5/accept",
+          "type": "application/json",
+          "class": ["comit-action"],
+          "fields": [
+            {
+              "name": "beta_ledger_refund_identity", "title": "Refund address on Ethereum", "type": "text",
```
LLFourn: Is type actually text here?

[edit] Oh I see it's the media type.

D4nte: yes, siren defines a list of acceptable values.

0003-siren-prototype-accept-decline.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+.The prototype
+[source,json]
+----
+{
+  "class": ["swaps"],
+  "entities": [
+    {
+      "class": ["swap"],
+      "rel": [],
```
D4nte: 🤔 I think this is unneeded in our context but I am not sure I understand its use:

> Defines the relationship of the sub-entity to its parent, per Web Linking (RFC5899).

0003-siren-prototype-accept-decline.adoc
```diff
@@ -0,0 +1,88 @@
...
+          "name": "accept",
+          "title": "Accept",
+          "method": "POST",
+          "href": "/swaps/rfc003/399e8ff5-9729-479e-aad8-49b03f8fc5d5/accept",
+          "type": "application/json",
+          "class": ["comit-action"],
+          "fields": [
+            {
+              "name": "beta_ledger_refund_identity", "title": "Refund address on Ethereum", "type": "text",
+              "class": ["address", "ethereum"]
```
thomaseizinger: :x: (to enforce discussion)

@LLFourn voiced some concerns at some point that this should maybe be a single value `ethereum-address` instead of two. What do you guys think?

0003-siren-prototype-accept-decline.adoc
thomaseizinger: Can just be `item` in our case.
See here: https://github.com/kevinswiber/siren/issues/17#issuecomment-26816414

thomaseizinger (changes requested):
Did you validate that it conforms to the schema? :)

0003-siren-prototype-accept-decline.adoc
D4nte: > Did you validate that it conforms to the schema? :)

Yes I configured CLion for that.

0003-siren-prototype-accept-decline.adoc
D4nte: yeah at first I was surprised to see `["address", "ethereum"]` in the draft as I was expecting `address-ethereum`.

I have no strong opinion on this. I think it makes sense to separate because at the end, when writing `address-ethereum` we are creating our own format to express the fact that it is an `address` on the `ethereum` network.
Having it separated makes slightly more sense to me.

thomaseizinger: The thing that makes separation a bit tricky is, how to you use these values in automated system?

You basically have to try and match against a known set of values to figure out, which classes are present because all you have is a list of strings. The logic would be like:

```
if classes.contains("ethereum") && classes.contains("address") {
    return metamask.getAddress()
}
```

Thoughts @LLFourn, @bonomat ?

D4nte: it'd probably be more nested than that:

```
if classes.contains("ethereum") {
  if classes.contains("address") {
    return metamask.getAddress()
  } else if classes.contain("broadcast-transaction") {
  // ...
  }
}
```

no?

thomaseizinger: I am not sure why any field would ever be tagged with a class `broadcast-transaction`?

Whether or not you nest it is a matter of taste I guess. I'd prefer not nesting it to keep things local and modular, i.e. detecting that a field is an ethereum address has little to do with detecting the field which represents the fee per byte on bitcoin. Nested `if` statements convey that requirements are connected, whereas if they are not, I'd prefer the code to not suggest that.

bonomat: Unfortunately siren does not give a lot of documentation about the thoughts behind this field besides: `Describes the nature of an entity's content based on the current representation. Possible values are implementation-dependent and should be documented. MUST be an array of strings. Optional` 

I googled quite a bit and there is not much usage of siren out there neither. 
However, I have the feeling that the *siren-way* of doing this would be indeed `["address", "ethereum"]`. 

The way how I would read it is that this field is of type `address` of type `ethereum`, or expressed in code: 
```
let ethereum_types = implementation_detail_types.get("ethereum")
let address_type = ethereum_types.get("address")
```

If we look at the [examples](https://github.com/kevinswiber/siren):
``` [...] "class": [ "info", "customer" ], [..] ```

this could be interpreted as: 
```
let customer_types = implementation_detail_types.get("customer")
let info_type = customer_types.get("info")
```

There can be some well known types which allow some kind of short forms and tell you directly what type it may be, e.g. 
```[...] "class": [ "items", "collection" ], [...]```

----

... on the other hand, to simplify things, I would go for the most specific version, i.e. `ethereum-address`  😬

Comment (D4nte) [reply 0003-siren-prototype-accept-decline.adoc]: Let's list the possible values to understand what would be the best solution. I think it does not matter in any case.

File: 0003-siren-prototype-accept-decline.adoc [hunk 4]
```diff
@@ -0,0 +1,88 @@
...
+        }
+      ],
+      "actions": [
+        {
+          "name": "accept",
+          "title": "Accept",
+          "method": "POST",
+          "href": "/swaps/rfc003/399e8ff5-9729-479e-aad8-49b03f8fc5d5/accept",
+          "type": "application/json",
+          "class": ["comit-action"],
```
D4nte: I agree with @bonomat this is useless.

0003-siren-prototype-accept-decline.adoc
thomaseizinger: @bonomat I don't really understand where you are going with your comment :thinking: 

From my understanding, `class` is a way of _tagging_ entities, fields and actions with domain-specific data. This serves the purpose of identifying and finding specific actions, fields or entities within the response.
The client would therefore:

- first take all actions in the response and look, which of them are tagged with `comit-action`
- those it would render as buttons in table row for the swap
- upon clicking an action, the client can inspect the `fields` of this action and, if present, render a form for all fields
- to enhance the user experience, it can search for **known** values in the `class` array to invoke special behaviour for some fields, like replacing the input field with an `EthereumAddressTextField` (which has support for metamask)

The question is therefore how "modular" we want this detection to be:

Do we want to repeat ourselves in defining single-token classes like:
- ethereum-address
- bitcoin-address
- bitcoin-fee
- ...

or do we want to split those tags into:

- bitcoin
- ethereum
- address
- fee
- ...

and look for combinations of them.

Splitting them up would make them more composable, which might be an advantage in the future.
On the other hand, the first option has the advantage of making invalid combinations of classes impossible.
However, invalid combinations are actually not so much of a issue since the client is only going to invoke special behaviour upon known combinations, hence and invalid combination will not break clients.

0003-siren-prototype-accept-decline.adoc
thomaseizinger: This is actually not useless at all!
You need a way of identifying, which actions should be rendered as buttons to the user. Assuming that all actions of a swap go in the _same_ list of choices is a bit dangerous IMO.

D4nte: I do not understand.
Can you please provide an example of 2 different actions where the sole differentiator for the API client is that one is a `"class": ["comit-action"]` and the other is something else?

0003-siren-prototype-accept-decline.adoc
D4nte: After coding it I prefer 2 classes to more easily transfer the responsibility of getting the data:

https://github.com/coblox/bobtimus/blob/f11d4c8c2730e950541b612e3fd9a583a307cf9a/src/action_handler.ts#L90

0003-siren-prototype-accept-decline.adoc
thomaseizinger: My concern is about forward compatibility. If we don't provide the class `comit-action`, clients will default to rendering all the actions of a swap next to each other. Currently, this is what we want. My thinking was that there may be SIREN actions at some point that are of a different kind other than the ones we currently have and thus we may want handle them differently from the current ones (render them in a different place for example).

Without a `comit-action` class, those will have to be explicitly filtered out whereas, with a `comit-action` class, the code that renders the current actions will always work, regardless of what we do in the future.

You could argue that we are never going to add any actions that need to be treated differently. Maybe that is the case, maybe not.

Thinking about it, it is probably too much premature optimization :)

0003-siren-prototype-accept-decline.adoc
bonomat: @thomaseizinger :
Along the lines this summarizes what I meant: 

> * to enhance the user experience, it can search for **known** values in the `class` array to invoke special behaviour for some fields, like replacing the input field with an `EthereumAddressTextField` (which has support for metamask)
...
> Splitting them up would make them more composable, which might be an advantage in the future.
> On the other hand, the first option has the advantage of making invalid combinations of classes impossible.

@D4nte 
> After coding it I prefer 2 classes to more easily transfer the responsibility of getting the data:

Well then let's go for this approach.

0003-siren-prototype-accept-decline.adoc
D4nte: I see your point and I don't think I have a strong opinion anymore :)

I am thinking that maybe a future action could be to do a 'POST' request on something else than the comit_node?

bonomat: @thomaseizinger : I'm not 100% sure I understand your thoughts:
if the client receives a POST action with some `fields`, then it should know that a form is required to collect the data (if it is an interactive client). 
I assume this should always be like that for all actions with some `fields`.
If there are no fields, then it's a simple button (e.g. the decline action). 

The way how I would interpret this is: if it is a `comit-action` render a `comit-logo` on top of the form/button/etc, if `ethereum-action`, render an `ethereum-logo`...

--> Happy to put it there for now .

0003-siren-prototype-accept-decline.adoc
thomaseizinger: Works for me!

0003-siren-prototype-accept-decline.adoc
D4nte: > @thomaseizinger : I'm not 100% sure I understand your thoughts:
> if the client receives a POST action with some `fields`, then it should know that a form is required to collect the data (if it is an interactive client).
> I assume this should always be like that for all actions with some `fields`.
> If there are no fields, then it's a simple button (e.g. the decline action).
> 
> The way how I would interpret this is: if it is a `comit-action` render a `comit-logo` on top of the form/button/etc, if `ethereum-action`, render an `ethereum-logo`...
> 
> --> Happy to put it there for now .

Best to hash it out face-to-face?

thomaseizinger: Discussed it offline - happy to remove `comit-action`.

0004-siren-prototype-accept-decline.adoc
```diff
@@ -0,0 +1,90 @@
+= Siren Prototype for comit-rs HTTP API - Accept/Decline actions
+Franck Royer <franck.royer@coblox.tech>
+:toc:
```
thomaseizinger, line 3: :x: Can we have the `toc` after the initial metadata (author, date and tracking issue)?

thomaseizinger (changes requested):

0004-siren-prototype-accept-decline.adoc
D4nte: No we can't as far as I understand/tried. `:toc:` keyword only works directly after the title.

bonomat: can confirm.. my mistake.

LLFourn (approved):

0004-siren-prototype-accept-decline.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+.The prototype
+[source,json]
+----
+{
+  "class": ["swaps"],
+  "entities": [
+    {
+      "class": ["swap"],
+      "rel": ["item"],
```
da-kami, line 29: Where does this point to? What is "item" ?

da-kami (approved):

0004-siren-prototype-accept-decline.adoc
thomaseizinger: Please have a look at the previous discussion in this PR:
https://github.com/comit-network/spikes/pull/7#discussion_r280271091

thomaseizinger (approved):
Good work! :)

PR #8, opened 2019-05-02
Convert template to asciidoc

D4nte:


thomaseizinger (approved):

template.adoc
```diff
@@ -1,14 +1,14 @@
-# [short title describing spike]
+= [short title describing spike]
 
 * Authors: [list everyone involved in the research]
```
thomaseizinger: I think asciidoc has a special notion for authors. Could use that.

template.adoc
```diff
@@ -1,14 +1,14 @@
...
 * Date: [YYYY-MM-DD when the spike outcome was last updated]
```
thomaseizinger: Same for this one :)

template.adoc
```diff
@@ -0,0 +1,15 @@
+= [short title describing spike]
+John Smith <john.smith@coblox.tech>; Fabien Miller <Fabien@coblox.tech>;
+
+* Author: {authors}
+* Date: {docdate}
```
D4nte: Note it's UTC :)

template.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+Tracking issue: [#?]link-to-corresponding-spike
+
```
bonomat, line 9: What do you think of adding a `:toc:` by default?

template.adoc
```diff
@@ -0,0 +1,15 @@
+= [short title describing spike]
+John Smith <john.smith@coblox.tech>; Fabien Miller <Fabien@coblox.tech>;
```
bonomat:
```suggestion
Satoshi Nakamoto <satoshi.nakamoto@coblox.tech>;
```

template.adoc
thomaseizinger: I think {docdate} should be set to something somewhere, otherwise this is going to change when you download the file etc. (The value is passed in by the renderer.)

Talked to @bonomat and we solved it like this:
https://github.com/comit-network/spikes/pull/6/files#diff-5991a65c40c76c0b0446b96e49e23868

thomaseizinger (changes requested):

template.adoc
```diff
@@ -0,0 +1,17 @@
+= [short title describing spike]
+Satoshi Nakamoto <satoshi.nakamoto@coblox.tech>;
+:toc:
```
thomaseizinger, line 3: As mentioned in the other document, I think having the ToC right at the front looks a bit off. I'd prefer having it after the meta-data (just before the content starts).

thomaseizinger (changes requested):

template.adoc
D4nte: No we can't as far as I understand/tried. `:toc:` keyword needs to be just after the title.

da-kami (approved):

template.adoc
thomaseizinger: Just checked the documentation it is possible but complicated :roll_eyes: 

http://asciidoc.org/userguide.html

This mentions the `toc-placement` attribute, which can be set to `manual` which allows you to use the `toc::[]` block macro to place the table of contents where you want.

D4nte: As you saw on my laptop, neither CLion asciidoc plugin nor GitHub asciidoc rendering support this `toc::[]` macro....

thomaseizinger: Sad that it doesn't work :(

thomaseizinger (approved):

PR #19, opened 2019-07-04
Write timestamp spike

tcharding:
As part of #1044 write a spike about timestamps.


tcharding:
@comit-network/rust-devs please be super nit-picky when reviewing this one.  Its my first spike, I'd like to get all the small details ironed out quickly as possible.  #mercylessreview Thanks.

thomaseizinger (dismissed):
Looks mostly fine to me. What I would like to see though is a recommendation of the author, which events we should now timestamp and whether those timestamps are the ones from the blockchain or when we received them.

Resolve this ambiguity and we are good to go!

tcharding:
> Looks mostly fine to me. What I would like to see though is a recommendation of the author

I did not include an opinion in this spike because I do not understand the motivation for timestamps.  The issue states only 'This just came up in a discussion: Events should have timestamps'.  Discussion meeting scheduled.

D4nte:
The aim of the timestamp is for the client to have an idea of the order of events and have the information. This would allow comit-i to display the events in the right order or it could allow a trading bot to analyse how much time a given peer is taking to do his part of a swap (ie, if a given peer is always taking a long time then he is likely to be using the option and hence the trading bot may decide to take action against this specific peer).

There is no possible "spoofing" in this scenario as we currently always assume that a client can trust its COMIT node.

As a base assumption for this spike, I would not expect comit_node to get any timestamp from another COMIT node. As you said, the comit_node should either make a timestamp by registering when a message arrives/is sent (from another COMIT node) or looking at the blockchain messages.

For this spike I'd expect to have the list of events, (swap request sent, acceptance of swap received, decline of swap received, transaction on blockchain seen/etc) and for each event have a list of the possible timestamps (ie, timestamp of when a message was sent vs timestamp of the transaction vs timestamp of the block etc) with a declaration of the risk for each (e.g., using timestamp of the block gives a precision of 10min etc).

tcharding:
Nice description @D4nte, thanks.  Now I better understand the spike.  I have cancelled the discussion meeting.

0010-timestamps.adoc
```diff
@@ -0,0 +1,78 @@
...
+ - representation i.e., wall clock time, seconds since epoch.
+
+== Events to timestamp
+
+Events that could potentially be timestamped.
+
+=== Events from Alice's perspective
+
+==== happy swap
+- swap accept received (fund available)
```
D4nte: I'd add `swap request sent`

0010-timestamps.adoc
```diff
@@ -0,0 +1,78 @@
...
+- swap decline received
+
+=== Events from Bob's perspective
+
+==== happy swap
+- swap request received
+- swap accept sent
+- Alpha ledger funded (fund available)
+- Beta ledger funded (refund available)
+- Beta asset redeemed (by Alice, redeem available)
```
D4nte: I'd add `beta/alpha asset refunded` for both Alice and Bob.

0010-timestamps.adoc
```diff
@@ -0,0 +1,78 @@
...
+- Alpha asset redeemed (by himself)
+
+====  decline swap
+- swap request received
+- swap decline sent
+
+== Suggestions
+
+I suggest we timestamp the events above using seconds since epoch.
```
D4nte, line 40: So if I understand correctly, your suggestion is to only timestamp the event from the moment `cnd` is made aware of them and ignore any timestamp coming from the blockchain.

It sounds reasonable but did you consider using timestamps from the blockchain?
E.g, when a `fund` transaction is seed, instead of saving the timestamp of when cnd is made aware, use the timestamp of the blockchain transaction (which is actually the timestamp of the block in which the transaction was confirmed).

Also, can we think about what would happen if we were to consider 6 confirmations for the `fund` event to be considered done?

We would have to assume that btsieve would return the transaction once all confirmations are done. So 1 hour after the transaction is broadcasted. Is that fine?

0010-timestamps.adoc
da-kami: Additionally I'd suggest renaming `happy swap` to something else that also depicts refund (e.g. `accepted swap`).
(`happy swap` sounds like the success scenario to me, which does not include refund from my point of view.)

0010-timestamps.adoc
da-kami: Question: What do we want to achieve with the timestamp?
If the sole purpose is, to put the event in order (so that comit-i can display them in the correct sequence) I think a timestamp by the comit-node would be sufficient. 

If we want to depict block confirmations in the timestamp then this would actually be a different timestamp for me. 
Do we have other use-cases besides the event order in comit-i that involve the timestamp?

tcharding: I limited the scope of these timestamps to just being of use to clients i.e. comit-i (stated in the document).  I do not think we should bother with blockchain times.

tcharding:
> So if I understand correctly, your suggestion is to only timestamp the event from the moment cnd is made aware of them and ignore any timestamp coming from the blockchain.

Yes

0010-timestamps.adoc
tcharding: Thank you, added.

0010-timestamps.adoc
D4nte: > I limited the scope of these timestamps to just being of use to clients i.e. comit-i (stated in the document). I do not think we should bother with blockchain times.

Can you add that to the spike then please?

0010-timestamps.adoc
tcharding: I intentionally left these out (hence name 'happy path').  Adding a section to the document stating the reasons.  You guys can ack/nack from there :)

tcharding: Rethought the above comment.  Added `refund`.

0010-timestamps.adoc
tcharding: Added 'I suggest that all timestamps are created by the cnd since they are only useful to the cnd I feel it adds no real benefit to get verified times from the blockchain.'

tcharding: Did patch with a this added as a 'suggestion'.

D4nte: Getting timestamp from the blockchain is not all about getting **verified** times, it's also about getting all events especially if we implement the confirmation feature.

Please note that I agree with your proposal, I just want to not leave any stones unturned.

0010-timestamps.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+== Events to timestamp
+
+Events that occur on a ledger have two times associated with them.
+The first is the time of the event occurs i.e., the ledger timestamp.
+The second is the time cnd is notified of the event (cnd is not guaranteed to be running when the event occurs).
+These events are marked below with '(ledger + cnd)', basically these are all the events except the initial COMIT messages.
+
+- swap request sent/received
+- --- swap decline sent/received
```
D4nte: what does `---` mean?

0010-timestamps.adoc
```diff
@@ -0,0 +1,73 @@
...
+- Beta ledger funded (redeem available) - (ledger + cnd)
+- --- Beta asset refunded - (ledger + cnd)
+- Beta asset redeemed  - (ledger + cnd)
+- Alpha asset redeemed - (ledger + cnd)
+- --- Alpha asset refunded - (ledger + cnd)
+
+== Conclusion
+
+In conclusion I recommend adding a timestamp generated by the cnd for each of the message events.
+For each notification from btsieve to cnd, recommend passing timestamp information (from the ledger) from btsieve to cnd and storing this information in cnd to make available to comt-i.
```
D4nte:
```suggestion
For each notification from btsieve to cnd, recommend passing timestamp information (from the ledger) from btsieve to cnd and storing this information in cnd to make available to comit-i.
```

0010-timestamps.adoc
```diff
@@ -0,0 +1,73 @@
...
+All timestamps should be made available to comit-i.
```
D4nte, line 74: So if I understand correctly you suggest we get the blockchain timestamp + cnd timestamp for all blockchain events?

0010-timestamps.adoc
tcharding: Damn, I was hoping that would be self evident.  it signals that all the other events form a sequence and the '---' events are branches.

0010-timestamps.adoc
tcharding: Yes

0010-timestamps.adoc
D4nte: huum, maybe find a better way?

0010-timestamps.adoc
D4nte: Could you make it more explicit?

D4nte (approved):
Approved, it'd be great if you could sort out the two cosmetic comments.

tcharding:
> Approved, it'd be great if you could sort out the two cosmetic comments.

The '---'?   Whats the other please @D4nte?

D4nte:
> > Approved, it'd be great if you could sort out the two cosmetic comments.
> 
> The '---'? Whats the other please @D4nte?

Describe your recommendation in a more explicit manner.

0010-timestamps.adoc
da-kami:
```suggestion
Both the cnd timestamps and the ledger timestamp should be made available to comit-i.
```
If this is what you mean I'd phrase it like this (or similar).

da-kami (approved):
Approved, but added one last nitpicky suggestion :)

PR #29, opened 2019-08-13
Write secure HTTP API spike

### LINKED ISSUE #477: How to secure the `cnd` HTTP API with HTTPS?
Answer the following questions:

- comit-i should be able to be served via HTTPS
- API should be accessible via HTTPS
- how to install the certificate?
- does it have to be self-signed? is there any funny workaround for localhost-only HTTPS APIs?

tcharding:
As defined in the issue [below], we would like to consider securing the HTTP API exposed by cnd.  'securing' is an ambiguous term, this spike is limited in scope to encrypting the traffic between cnd and comit-i.

Fixes: https://github.com/comit-network/comit-rs/issues/477

### DIFF
--- a/0013-secure-http-api.adoc
+++ b/0013-secure-http-api.adoc
@@ -1,5 +1,4 @@
 = Securing the `cnd` HTTP API
-
 Tobin C. Harding <tobin.harding@coblox.tech>
 :toc:
 :revdate: 2019-08-13
--- a/0013-secure-http-api.adoc
+++ b/0013-secure-http-api.adoc
@@ -52,34 +52,94 @@ The industry standard for encrypting traffic over a network is Transport Layer S
 * Replay protection - not needed, replay attack vector currently non-existent for cnd.
 * Sequence protection - no obvious need.
 
-There is a usability concern that for clients that use a webbrowser e.g. `comit-i` how we sign the certificate will effect the user experience.
-A self signed cert will cause a security warning to be displayed by modern browsers.
-Installing a CA to counter the issue will require root privileges, something that we do not wish to require.
+=== Certificates
+
+TLS requires a certificate (private/public keypair) to be preset on the server i.e., cnd.
+Note, this negatively impacts our quarterly OKR #48 (Make COMIT easily accessible).
+
+There are multiple solutions to getting a certificate onto the machine running cnd, each with different pros/cons.
+
+==== Signed certificate by trusted CA
+
+Use case: running cnd on an internet facing host with a registered domain.
+Purchase a certificate signed by a trusted CA.
+
+.pros
+* This is how it would need to be done if cnd is offered as a service i.e., to mobile clients
+* Industry standard
+
+.cons
+* Resource intensive, money and time.
+* Unnecessary for use cases where cnd is on the same machine as the client
+* Not suitable for development
+
+==== Signed certificate by private CA
+
+Use case: running cnd on same machine as the client.
+Create a private CA and configure the host machine to trust certificates signed by the private CA.
+
+.pros
+* Free
+* Makes using the client with HTTPS seamless
+
+.cons
+* Requires root access to the host machine
+* Complex if done manually, however there is a tool (untested) that claims to simplify this: https://blog.filippo.io/mkcert-valid-https-certificates-for-localhost/
+* Operator of cnd still has to create and sign certificates for use by cnd i.e., increases complexity
+
+=== Self signed certificate
+
+Use case: running cnd on same machine as the client.
+Create a self signed certificate.
+
+.pros
+* Free
+* Within the realm of typical developer knowledge
+
+.cons
+* Adds complexity to the set up of cnd
+* Negatively impacts usability, browser security pop up because of untrusted certificate
 
 == Implementing TLS
+
+There are two aspects to encrypting traffic between cnd to the client.
+1. The HTTP API exposed by cnd needs to be encrypted using TLS.
+2. *Our* client implementation, `comit-i` needs to be served by cnd using TLS.
 
 === cnd
 
 Adding TLS to cnd should be straightforward.
-TLS is supported by Warp (the web server framework we use).
-The only real consideration is the TLS keypair (key and certificate) and selfsigning the certificate.
-We can provide a default keypair so that cnd works out of the box with TLS.
 
-.This can be done in a few ways:
-* Put the keys in the repository - bad, private keys should be private.
-* Write a standalone tool to generate a new key.
-* Add code to cnd to generate a new keypair if one is not found i.e. in the default place or via a config option `--key=/path/to/key`and `--cert=/path/to/cert`.
+There are two parts to this:
+1. Exposing the HTTP API vial TLS (so clients can connect to cnd using TLS)
+2. Serving `comit-i` via TLS (so the browser can use `https://` addresses)
+
+Since both of these are done by calls to Warp they can both be done in the same manner.
+TLS is supported by Warp (the web server framework we use).
+Requires adding a field in the config file and command line options e.g., `--tls-key=/path/to/key` and `--tls-cert=/path/to/cert`.
+Adds an extra initial setup step for the user of cnd to ensure valid certificate is available on the host machine.
 
 === comit-i
 
-Request content for this section to be written by team member who is familiar with `comit-i` and/or React.
+`comit-i` must consume the HTTP API exposed by cnd using TLS
+I was unable to find any information on the internet on how to do this.
 
 == Recommendation
+
+The cnd stuff is straightforward.
 
 .Recommendations are as follows:
 * Implement TLS for cnd using Warp's TLS support.
 ** Enable TLS by default, add option `--no-tls` to disable it.
 ** Look for keypair in well known location, configurable via configuration file and command line options.
+* Leave generation of valid certificate up to the user of cnd but add documentation/hints on how to do this and what options are available (based on information in this spike).
+
+As for `comit-i` my findings were less positive.
+Recommend splitting `comit-i` up into separate parts for the front end and back end and re-writing the back end in Rust.
+Rust has a library `rust-native-tls` which is an abstraction over platform-specific TLS implementations and will facilitate connecting to cnd using TLS.
+
+.Failed research question:
+* Connect to cnd from `comit-i` using TLS via Typescript or Javascript library.
 
 == Reference
 
--- a/0013-secure-http-api.adoc
+++ b/0013-secure-http-api.adoc
@@ -4,8 +4,8 @@ Tobin C. Harding <tobin.harding@coblox.tech>
 :toc:
 :revdate: 2019-08-13
 
-* Authors(s): {authors}
-* Date: {revdate}
+* Authors(s): {authors} +
+* Date: {revdate} +
 
 Issue: https://github.com/comit-network/comit-rs/issues/897[#477]
 
--- a/0013-secure-http-api.adoc
+++ b/0013-secure-http-api.adoc
@@ -26,14 +26,14 @@ The requirement for encryption is self evident and is the primary focus of this 
 The topic of authentication is not so clear cut.
 Authentication is actually out of scope for this spike but briefly considering here for completeness.
 
-Client authentication refers to three things:
+.Client authentication refers to three things:
 * Identity – who claims to be making an API request?
 * Authentication – are they really who they say they are?
 * Authorization – are they allowed to do what they are trying to do?
 
 At the moment only the first two are of interest to us, all users are allowed to do anything.
 
-Considerations:
+.Considerations:
 * API key
 * username/password
 * Basic authentication with TLS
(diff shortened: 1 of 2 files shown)


bonomat (changes requested):
Thanks for the summary. 

I think you can go one step further and propose a solution to how we can secure cnd's API exactly. I'd say *not* doing this is not an option if we want to write *best practice software* as you state at the beginning.

tcharding:
> I'd say _not_ doing this is not an option if we want to write _best practice software_ as you state at the beginning.

Fair point, will do further research.

tcharding:
Ok, spike new and revised - I'm happy with my conclusions and that someone could now go ahead and implement TLS as recommended in this spike without having to do further research.  Please say if this view is incorrect - I'd really like to get this whole spike writing thing under my belt.

thomaseizinger (commented):
The problem with self-signed certs is that the browser will not trust them and display a giant warning to the user that what he tries to do is unsafe. IMO, this is not an acceptable solution.

If we want to use a self-signed certificate, I think we would need to install it as a CA on the computer (root privileges required) which seems to be bad.

This documentation here on letsencrypt.org goes into more detail for this problem: https://letsencrypt.org/docs/certificates-for-localhost/

Regarding the spike: I think the approach is good but don't think it will be possible to have a good solution to this if `comit-it` is out-of-scope :)

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,85 @@
+= Securing the `cnd` HTTP API
+
```
D4nte:
```suggestion
```
Must not have a new line for ascii doc to pick up the author line.

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,85 @@
...
+Tobin C. Harding <tobin.harding@coblox.tech>
+:toc:
+:revdate: 2019-08-13
+
+* Authors(s): {authors}
```
D4nte:
```suggestion
* Authors(s): {authors} +
* Date: {revdate} +
```
There is a '+' in the template, not sure what was the reason but pretty sure you need to keep it

D4nte (commented):
@tcharding can you do a pass on the asciidoc format? Just check the file from GitHub and correct what does not render well (new line is needed before a list for example).

0013-secure-http-api.adoc
thomaseizinger: Is needed for correct rendering.

tcharding:
> @tcharding can you do a pass on the asciidoc format? Just check the file from GitHub and correct what does not render well (new line is needed before a list for example).

My bad, thanks for poking me.

tcharding:
> The problem with self-signed certs is that the browser will not trust them and display a giant warning to the user that what he tries to do is unsafe. IMO, this is not an acceptable solution.

Yes, I did not include that in the spike I assumed it was common knowledge and acceptable.

> If we want to use a self-signed certificate, I think we would need to install it as a CA on the computer (root privileges required) which seems to be bad.

This is worse IMO than just expecting the user to click through the warning - since our target are devs I feel it is safe to assume they know what doing that means and since they most likely started the cnd themselves this will not be a surprise.  Should this discussion go into the spike? 

> This documentation here on letsencrypt.org goes into more detail for this problem: [letsencrypt.org/docs/certificates-for-localhost](https://letsencrypt.org/docs/certificates-for-localhost/)
> 
> Regarding the spike: I think the approach is good but don't think it will be possible to have a good solution to this if `comit-it` is out-of-scope :)

Oh, does it read like `comit-i` is out of scope.  That was not my intention, I was hoping someone with front end knowledge would speak up.  Perhaps a better method of getting this information would be for me to actively ask :)

0013-secure-http-api.adoc
bonomat: It's to get a new line after `* Authors(s): {authors}`
https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#breaks

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,85 @@
...
+The only real consideration is the TLS keypair (key and certificate) and selfsigning the certificate.
+We can provide a default keypair so that cnd works out of the box with TLS.
+This can be done in a few ways:
+* Put the keys in the repository - bad, private keys should be private.
+* Write a standalone tool to generate a new key.
+* Add code to cnd to generate a new keypair if one is not found i.e. in the default place or via a config option `--key=/path/to/key`and `--cert=/path/to/cert`.
+
+=== comit-i
+
+Request content for this section to be written by team member who is familiar with `comit-i` and/or React.
```
bonomat: In order to get some knowledge around comit-i, please pair up with someone and write it together.

D4nte:
> This is worse IMO than just expecting the user to click through the warning - since our target are devs I feel it is safe to assume they know what doing that means and since they most likely started the cnd themselves this will not be a surprise. Should this discussion go into the spike?

Our target is devs that build apps used by users. In this case, the final user would still need to click through the warning if the app is the webapp which is not good.

_Also, please avoid starting discussions on the main thread for spikes, makes it hard to follow_

0013-secure-http-api.adoc
tcharding: > In order to get some knowledge around comit-i, please pair up with someone and write it together.

Ok

tcharding:
> Also, please avoid starting discussions on the main thread for spikes, makes it hard to follow

Where should these discussions be please @D4nte?  On the original issue?

thomaseizinger:
> Also, please avoid starting discussions on the main thread for spikes, makes it hard to follow

Probably my bad because I voiced the concern in the general review, with does not allow to create threads.

@tcharding 

Thanks for amending the spike with those concerns! How are we dealing with that problem now?

Didn't read it yet but maybe this HN thread has some insights: https://news.ycombinator.com/item?id=17358339

tcharding:
> Didn't read it yet but maybe this HN thread has some insights: news.ycombinator.com/item?id=17358339

Good link, thanks.  According to the blog post linked (indirectly) above; we may be able to get around using self signed certificates by using `127.0.0.1` instead of `localhost`.  Turns out modern browsers treat these addresses differently and the IP version is considered ["potentially trustworthy"](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy).

D4nte:
> > Also, please avoid starting discussions on the main thread for spikes, makes it hard to follow
> 
> Where should these discussions be please @D4nte? On the original issue?

In the "Files changed" tab, on the line concerned by the comment.

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,152 @@
...
+.What TLS gives us and whether we need it:
+* Authentication of cnd - needed, allows client to verify it is connecting to the correct cnd.
+* Data confidentiality - needed, protects against attacker learning the secret.
+* Date integrity - needed, protects against modification of the data i.e. changing the redeem address.
+* Replay protection - not needed, replay attack vector currently non-existent for cnd.
+* Sequence protection - no obvious need.
+
+=== Certificates
+
+TLS requires a certificate (private/public keypair) to be preset on the server i.e., cnd.
```
D4nte:
```suggestion
TLS requires a certificate (private/public keypair) to be present on the server i.e., cnd.
```

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,152 @@
...
+There are two aspects to encrypting traffic between cnd to the client.
+1. The HTTP API exposed by cnd needs to be encrypted using TLS.
+2. *Our* client implementation, `comit-i` needs to be served by cnd using TLS.
+
+=== cnd
+
+Adding TLS to cnd should be straightforward.
+
+There are two parts to this:
+1. Exposing the HTTP API vial TLS (so clients can connect to cnd using TLS)
```
D4nte, line 120:
```suggestion
1. Exposing the HTTP API over TLS (so clients can connect to cnd using TLS)
```

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,152 @@
...
+2. Serving `comit-i` via TLS (so the browser can use `https://` addresses)
```
D4nte, line 121:
```suggestion
2. Serving `comit-i` over TLS (so the browser can use `https://` addresses)
```

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+Since both of these are done by calls to Warp they can both be done in the same manner.
+TLS is supported by Warp (the web server framework we use).
+Requires adding a field in the config file and command line options e.g., `--tls-key=/path/to/key` and `--tls-cert=/path/to/cert`.
+Adds an extra initial setup step for the user of cnd to ensure valid certificate is available on the host machine.
+
+=== comit-i
+
+`comit-i` must consume the HTTP API exposed by cnd using TLS
+I was unable to find any information on the internet on how to do this.
```
D4nte: I assume you just update the cnd REST API end point from `http` to `https`.

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,152 @@
...
+The cnd stuff is straightforward.
+
+.Recommendations are as follows:
+* Implement TLS for cnd using Warp's TLS support.
+** Enable TLS by default, add option `--no-tls` to disable it.
+** Look for keypair in well known location, configurable via configuration file and command line options.
+* Leave generation of valid certificate up to the user of cnd but add documentation/hints on how to do this and what options are available (based on information in this spike).
+
+As for `comit-i` my findings were less positive.
+Recommend splitting `comit-i` up into separate parts for the front end and back end and re-writing the back end in Rust.
```
D4nte: I don't understand why you would need that.

0013-secure-http-api.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+Use case: connecting to `comit-i` with the machine running cnd.
+Create a self signed certificate.
+
+.pros
+* Free
+* Within the realm of typical developer knowledge
+
+.cons
+* Adds complexity to the set up of cnd
```
D4nte, line 100: Why not:
- cnd generates certificates when starting up if not yet existing and if `webgui` is activated and `webgui.tls` (new option) is activated. This will be a CA, Public Cert and Private Cert
- User is told to install the freshly generated CA

That's it.

Feel free to at `lnd` as it does generate certificates for its gRPC API on the spot as described above (it does not generate a CA though, only a `tls.cert` and `tls.key`).

0013-secure-http-api.adoc
```diff
@@ -52,34 +52,98 @@ The industry standard for encrypting traffic over a network is Transport Layer S
...
+The cnd stuff is straightforward.
+
 .Recommendations are as follows:
 * Implement TLS for cnd using Warp's TLS support.
 ** Enable TLS by default, add option `--no-tls` to disable it.
 ** Look for keypair in well known location, configurable via configuration file and command line options.
+* Leave generation of valid certificate up to the user of cnd but add documentation/hints on how to do this and what options are available (based on information in this spike).
+
+As for `comit-i` my findings were less positive.
+Recommend splitting `comit-i` up into separate parts for the front end and back end and re-writing the back end in Rust.
```
thomaseizinger: Is your idea that we can then serve `comit-i` on `http` again?
I am afraid, this does not really help us :(
One of the reasons why it would be useful to serve `comit-i` via HTTPS is to make integrations with a Ledger Nano S possible (the browser only allows you to talk to a USB device on HTTPS)

thomaseizinger (approved):
I am starting to like this :)

It looks like our best option is to on-demand create a certificate for cnd if the user wants to have TLS for the API.

0013-secure-http-api.adoc
tcharding: Unless I'm misunderstanding the architechture, I think it is more complicated than that.  `comit-i` consumes the REST API via typescript code, not via the web browser so `comit-i` has to do the TLS handshake.

0013-secure-http-api.adoc
tcharding: Because of my comment above, if it is correct.

0013-secure-http-api.adoc
tcharding: I think I get what you are saying but am confused by the terminology - I'm only just across all the TLS certificate terms as is.  Either way we still either make the user click through security warnings or require root to install the CA.

0013-secure-http-api.adoc
tcharding: The idea is to serve `comit-i` using HTTPS.

0013-secure-http-api.adoc
tcharding: I do not think that `cnd` should create the certificate.  I think we should stick to the UNIX principle of doing one thing.  The user can generate the certificate.

tcharding:
> It looks like our best option is to on-demand create a certificate for cnd if the user wants to have TLS for the API.

As commented above, I don't think we should be creating the certificate in `cnd`, I feel that we should leave it up to the user to do so.

thomaseizinger (approved):
Thanks for adding the final notes @luckysori !

FYI @tcharding, the final notes are the conclusion of our discussion this morning :)

bonomat (approved):

PR #31, opened 2019-08-15
Fix ascii doc rendering

tcharding:
Recent spikes were not merged without the author (me) checking the rendering.


tcharding:
Attempted to put the issue link number in a separate patch to prevent confusion while reviewing.  You may not notice this of not viewing individual patches during review.  Thanks.

da-kami (approved):

PR #33, opened 2019-08-16
Add spike for proposed solution to resume swap after restart

### LINKED ISSUE #1061: Resume swap after restart of cnd
Assumptions:
- Btsieve is still running
- Btsieve still holds the previous queries with the same ids

Notes regarding the current state machine:
Currently, the state machine is based on futures and advanced "by itself" as these futures resolve. This is suboptimal for a number of reasons:
- It is hard to restart from an intermediate state/mid-swap (this ticket)
- It forces the module providing the state machine to depend on many dependencies (making it hard/impossible) to provide it in the form of a library
- It is very big in terms LoC (an approach not based on futures is probably much smaller)

Notes regarding the ticket:
With a different implementation of the swap protocol state machine, this ticket might be much easier because no "advancing" whatsoever is needed.

thomaseizinger:
Resolves https://github.com/comit-network/comit-rs/issues/1061.

### DIFF
--- a/0014-resume-swaps-after-restart.adoc
+++ b/0014-resume-swaps-after-restart.adoc
@@ -107,6 +107,7 @@ index eb46c727..0aafc5c2 100644
 
 As a result, we don't care if the query already exists on btsieve by the time we issue this request.
 Hence, we also don't need to store the URL or anything else in order to resume a swap.
-All we need to do is re-instantiate the state machine in the current state (retrieved from the database) and upon the first poll, this code will reconstruct the future with the existing query on btsieve.
+All we need to do is re-instantiate the state machine in the `Start` state with the initial value.
+Upon the first poll, this code will reconstruct the future with the existing query on btsieve and fast forward to the current state.
 
-Shoutout to @da-kami for coming up with the initial idea for this!
+Shoutout to @da-kami for coming up with the idea of using deterministic query IDs on btsieve!
--- a/0014-resume-swaps-after-restart.adoc
+++ b/0014-resume-swaps-after-restart.adoc
@@ -105,6 +105,8 @@ index eb46c727..0aafc5c2 100644
              .and_then(move |tx| {
 ----
 
-With this change, we don't care if the query already exists on btsieve by the time we issue this request.
+As a result, we don't care if the query already exists on btsieve by the time we issue this request.
 Hence, we also don't need to store the URL or anything else in order to resume a swap.
 All we need to do is re-instantiate the state machine in the current state (retrieved from the database) and upon the first poll, this code will reconstruct the future with the existing query on btsieve.
+
+Shoutout to @da-kami for coming up with the initial idea for this!
--- a/0014-resume-swaps-after-restart.adoc
+++ b/0014-resume-swaps-after-restart.adoc
@@ -1,0 +1,110 @@
+= Resume a SWAP after restart of cnd
+Daniel Karzel <daniel.karzel@coblox.tech>; Thomas Eizinger <thomas@coblox.tech>;
+:toc:
+:revdate: 2019-08-15
+
+NOTE: Author: {authors} +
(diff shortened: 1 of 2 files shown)


luckysori (approved):
Looks good to my naive eyes.

D4nte (changes requested):
Please Update README to add reference to the spike.

D4nte:
Looks good! State Machine Fast Forward ⏩

0014-resume-swaps-after-restart.adoc
```diff
@@ -0,0 +1,110 @@
...
+-            .create(BitcoinQuery::deploy_htlc(htlc_params.compute_address()))
++            .create(identifier, BitcoinQuery::deploy_htlc(htlc_params.compute_address()))
+             .and_then(move |query_id| query_bitcoin.transaction_first_result(&query_id))
+             .map_err(rfc003::Error::Btsieve)
+             .and_then(move |tx| {
+----
+
+With this change, we don't care if the query already exists on btsieve by the time we issue this request.
+Hence, we also don't need to store the URL or anything else in order to resume a swap.
+All we need to do is re-instantiate the state machine in the current state (retrieved from the database) and upon the first poll, this code will reconstruct the future with the existing query on btsieve.
```
bonomat: We should also think about if we have _enough_ data to recover easily and how to restore from it in a smart way? 

For example, think about this case: 
Bob's cnd restarts just before he redeems (redeem hasn't happened). 
What do we store in the DB? Is it only the data you mention above?

- Pair of ledgers
- Pair of assets
- Secret/SecretHash
- Expiry times
- Identities

Afaik, with only this information available, we will need to create (at least) 3 queries and wait for btsieve to find the results (meaning querying the blockchain again if they were deleted):
* query for alpha funded
* query for beta funded
* query for alpha redeemed

However, if we store the past events/results in the database, (e.g. alpha redeemed was seen prior the restart) such as 
* alpha funded transaction
* beta funded transaction
* alpha redeemed transaction 
Then we could recover faster/nicer.

What do you think?

bonomat (changes requested):
Cool, sounds like an easy fix :D 

2 more comments: 
* one in the code
* can you update README.md as well please?

0014-resume-swaps-after-restart.adoc
thomaseizinger: > (meaning querying the blockchain again if they were deleted):

The scope of the spike (in the ticket) states that we can assume that the queries are still available on btsieve.

> However, if we store the past events/results in the database, (e.g. alpha redeemed was seen prior the restart)

I would try to avoid this as much as possible and keep the whole process idempotent because I think it is easier to reason about.

> Then we could recover faster/nicer.

Haven't done any performance testing of the proposed solution but I'd be surprised if this takes very long. In the end, it is a couple of HTTP requests to btsieve which retrieves the existing results from the database. That should be pretty fast :)

0014-resume-swaps-after-restart.adoc
```diff
@@ -105,6 +105,8 @@ index eb46c727..0aafc5c2 100644
              .and_then(move |tx| {
 ----
 
-With this change, we don't care if the query already exists on btsieve by the time we issue this request.
+As a result, we don't care if the query already exists on btsieve by the time we issue this request.
 Hence, we also don't need to store the URL or anything else in order to resume a swap.
 All we need to do is re-instantiate the state machine in the current state (retrieved from the database) and upon the first poll, this code will reconstruct the future with the existing query on btsieve.
+
+Shoutout to @da-kami for coming up with the initial idea for this!
```
D4nte: Team effort dude!

D4nte (approved):
As discussed, please explicitly mention that the state machine will be start as `Start` in the recovery.

bonomat (approved):

PR #36, opened 2019-08-23
Correct the tag action as we do not have a develop branch

D4nte:
As discussed, if we were to tag on master then another PR could get into the release by mistake.

Tagging on master works for the standard Git Flow because only branch releases are merged in master, all other PRs are merged in a _develop_ branch.

However we do not have and dot not want a _develop_ branch for now, hence tagging on the release branch is the proposed solution.

### DIFF
--- a/0012-release-strategy.adoc
+++ b/0012-release-strategy.adoc
@@ -488,41 +488,47 @@ I recommend 5. (the pros and cons are already in the table) :)
 
 What should actually happen when releasing? Let's list below (I love tables)
 
-[cols="1,2"]
+[cols="1,1,2"]
 |===
-|What |Why
+|Step |What |Why
 
+|#1
 | Open a release issue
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):

0012-release-strategy.adoc
```diff
@@ -488,41 +488,47 @@ I recommend 5. (the pros and cons are already in the table) :)
...
+| The CI hook creates binaries and attach them to GitHub Release
+| No action from releaser
 
+|#8
 | Tweet about the release
 | Why creating a changelog if nobody reads it
 
+|#9
 | Mention it in the next blog post
 | Ditto.
```
luckysori, line 534: :wrench: Otherwise the next item won't appear on the table.
```suggestion

|#10
```

luckysori (approved):

PR #37, opened 2019-08-28
Improve feedback on failure of Ethereum HTLCs

### LINKED ISSUE #62: How should the Ethereum HTLC react to invalid invocations?
There are two ways of invoking the Ethereum HTLC in an invalid way:

- with the wrong secret
- without the secret but before the expiry

In both cases, we currently `return` from the contract without doing anything. This makes it hard to observe, whether or not the invocation was actually successful. From an EVM perspective, it executes successfully (i.e. without any errors)
Such an observation would be useful in numerous ways:
1. In our e2e tests to check if the refund actually worked or if we have to invoke it again
2. Metamask may show something different if the user invokes a refund too early. Showing a successful transaction if it actually was not successful is pretty misleading.

---

Ethereum has the `revert` instruction to halt execution of a contract and rollback any changes.

This spike is about investigating, what the best way is to _fail_ the invoking transaction of the HTLC.

D4nte:
Resolves comit-network/RFCs#62.

### DIFF
--- a/0015-ethereum-htlc-graceful-exit.adoc
+++ b/0015-ethereum-htlc-graceful-exit.adoc
@@ -82,7 +82,7 @@ The list of closure instructions (`0xf0` range):
 The [current Ether Basic HTLC](https://github.com/comit-network/RFCs/blob/master/RFC-007-SWAP-Basic-Ether.md#contract) returns if there an issue:
 
 .Current Ether Basic HTLC
-[source,asm]
+[source]
 ----
 {
     // Load received secret size
@@ -214,7 +214,7 @@ For this example, we propose 2 error messages:
 - `tooEarly`: if no secret is provided and expiry has not yet happened
 
 .Proposed new Ether Basic HTLC using revert
-[source,asm]
+[source]
 ----
 {
     // Load received secret size
--- a/0015-ethereum-htlc-graceful-exit.adoc
+++ b/0015-ethereum-htlc-graceful-exit.adoc
@@ -57,7 +57,7 @@ As per https://github.com/ethereum/EIPs/blob/master/EIPS/eip-140.md[EIP-140]:
 
 Which means that the REVERT instructions gives us the benefits we are seeking:
 
-- Stop the execution: _The effect of REVERT is that execution is aborted, considered as failed, and state changes are rolled back._
+- Stop the execution: _The effect of REVERT is that execution is aborted, *considered as failed*, and state changes are rolled back._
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):
Sweet! Good work :)

0015-ethereum-htlc-graceful-exit.adoc
```diff
@@ -0,0 +1,310 @@
...
+
+
+
+=== Improving logging
+
+Whether we use `return` or `revert`, we could improve the logging by returning or `log`-ging data at the end of a failed execution.
+
+We need to consider whether to
+
+- Use the same message for A, B & C.
```
luckysori: :upside_down_face: Slightly confusing to talk about `A`, `B` and `C` before they're introduced. Also the list that introduces them below is actually rendered as a numbered list so it looks like they're never introduced :stuck_out_tongue:

luckysori (approved):
Very thorough and clear :smile:

0015-ethereum-htlc-graceful-exit.adoc
luckysori: Aug28::163005

D4nte: haha yeah I re-ordered few things there and there. thanks.

0015-ethereum-htlc-graceful-exit.adoc
```diff
@@ -0,0 +1,316 @@
...
+    // Jump to redeem if hashes match
+    redeem
+    jumpi
+
+// Exit if it does not match
+exit_secret:
+    // Load Error Message "WRONG SECRET"
+    log1(0, 32, <keccak256("wrongSecret()")>)
+    // Exit
+    revert(0, 0)
```
bonomat, line 278: Since `revert` allows you to return something, did you consider something like this? Or did you exclude this because revert would need to load something from memory?
`revert("tooEarly()")`

bonomat (approved):
Well done, nicely written. Added a question but I guess I know why you excluded that option :)

0015-ethereum-htlc-graceful-exit.adoc
thomaseizinger: I think @D4nte mentions somewhere that only _other_ contracts can read this return data, i.e. if you call the contract with a transaction, you don't have access to this. So it looks like we should emit a log to state what is going on!

D4nte: Excluded because only another contract using such function can access the returned data.
Returned data can be access from web3 if, and only if, it's a readonly function that can be `call`ed. See the `Return vs Log error message` section for more details.

PR #43, opened 2019-12-13
Initial draft of new domain model spike

### LINKED ISSUE #1708: Towards a richer domain model
## Problem

Currently, several of the core types in our codebase come from external libraries:

- bitcoin::Amount,
- web3::Address,

This is problematic in several ways:

1. The names of these types are fixed
2. We cannot change the API
3. We cannot implement foreign traits on those types

We can workaround some of these problems (f.e. (3)) with solutions like generic newtypes. However, I see the core problem in (1). Our domain model is in-consistent in several ways (Amount vs Quantity for example).

## Goal

Have a richer domain model that models the domain as accurately as possible, both in terms of type names and their APIs.

## Recommendation

Introduce our own types for all crucial concepts of our domain:

- amounts
- identities (public keys & addresses)

We don't have to implement all of the logic of those types ourselves. They can wrap library types to get access to certain functionality.
As part of this ticket, we should also trash the `ethereum_support` crate as most of that functionality is only used inside `cnd`.

Tagging this one as a spike because it is mostly a design task. The choices in names are important here and should be discussed and documented. Creating a spike seems like a good fit!

thomaseizinger:
Resolves https://github.com/comit-network/comit-rs/issues/1708.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ List of spike outcomes for COMIT network.
 - [spike-0014](0014-resume-swaps-after-restart.adoc) - Resume swaps after restart
 - [spike-0015](0015-ethereum-htlc-graceful-exit.adoc) - Graceful exit of the Ethereum HTLC for incorrect invocations
 - [spike-0016](0016-how-to-document-cnd-http-api.adoc) - How to document cnd's HTTP API
+- [spike-0017](0017-negotiation-and-execution-protocol.adoc) - Separating negotiation and execution phase
 - [spike-0018](0018-async-stack.adoc) - Decide on an async stack
+- [spike-0019](0019-domain-model.adoc) - Towards a richer domain model in cnd
 
 For new spike outcomes, please use [template.adoc](template.adoc) as basis.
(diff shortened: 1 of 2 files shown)


tcharding (approved):
Good thoughts @thomaseizinger.

README.md
```diff
@@ -21,5 +21,7 @@ List of spike outcomes for COMIT network.
 - [spike-0014](0014-resume-swaps-after-restart.adoc) - Resume swaps after restart
 - [spike-0015](0015-ethereum-htlc-graceful-exit.adoc) - Graceful exit of the Ethereum HTLC for incorrect invocations
 - [spike-0016](0016-how-to-document-cnd-http-api.adoc) - How to document cnd's HTTP API
+- [spike-0017](0017-negotiation-and-execution-protocol.adoc) - Separating negotiation and execution phase
+- [spike-0018](0016-how-to-document-cnd-http-api.adoc) - How to document cnd's HTTP API
```
luckysori:
```suggestion
```

luckysori (commented):
On board with all your thoughts and suggestions 👍

luckysori (approved):

D4nte (approved):
LGTM

PR #45, opened 2019-12-23
Write logging spike

tcharding:
This spike makes researches how we may go about improving our current logging.  There is a companion PR: https://github.com/comit-network/comit-rs/pull/1786 that implements the initial spike suggestions.

And POC for adding tracing to cnd: https://github.com/comit-network/comit-rs/pull/1888

### DIFF
--- a/0018-logging.adoc
+++ b/0018-logging.adoc
@@ -1,0 +1,94 @@
+= [Overhaul comit-rs Logging]
+Tobin C. Harding <tobin.harding@coblox.tech>;
+:toc:
+:revdate: 2019-12-23
+
+NOTE: Author: {authors} +
+Date: {revdate} +
+Tracking issue: https://github.com/comit-network/comit-rs/issues/487[#487]
+
+== Context
+
+At the moment, there is no consistency in what is logged where and with what level of detail.
+Debugging cnd by using the logs is painful at best and impractical at worst.
+
+This spike is about investigating, how the current logs could be improved and proposing an overall logging strategy that makes the logs actually useful.
+
+There are a few problems that need addressing, including but not limited to:
+
+* Log levels - guidelines on when to log and at what level
+* Useful context - we need useful context, what is useful should be defined by the team since it comes from experience debugging cnd.
+* Only useful output i.e., no dumping u8 arrays in the logs.
+* Async codebase - since we have an asynchronous code base logging may be inherently the wrong approach, we will also consider tracing.
+* Dynamically controlling log level
+* Controlling the log level of libraries vs cnd logs i.e., 
+
+The output of this spike should be proposed changes to the current logging libraries (i.e., what libraries we use to log) and also some guidelines so the whole team knows when and where to add log statements.
+
+== Research
+
+=== Log Levels ===
+
+
+This is actually the easiest thing to solve, using correct log levels is pretty standard dev practice, the internet more or less has consensus on the following log levels:
+
+Copying from the following site: https://reflectoring.io/logging-levels/
+
+* error: should only be used when the application really is in trouble. Users are being affected without having a way to work around the issue.
+* warn: should be used when something bad happened, but the application still has the chance to heal itself or the issue can wait a day or two to be fixed.
+* info: should be used to document state changes in the application or some entity within the application.
+* debug: should be used to log any information that helps us identify what went wrong on DEBUG level.
+* trace: log all information that helps us to trace the flow of the application.
+
+=== Useful Context ===
+
+Useful context is not that easy to solve due to the async nature of cnd.
+Typically one can follow the sequence of log messages to get more context than is available within the scope of the log statement.
+The solution to this is most likely to use a tracing library instead of logging, see below for more on tracing.
+
+Related to context is useful output in the log messages i.e., don't dump byte slices using "{:?}".
+
+Suggestions to remedy this include:
+* Implement custom `Debug` for any objects that we would like to print within logs.
+* Add a `Log` trait for any objects that we wish to log and call the method when logging i.e., `log::info("Got: {}", foo.log())`
+
+=== Controlling Log Levels ===
+
+Currently we can control the base log level within the config file.  More fine grained control requires code changes in `cnd/src/logging/initialize.rs`.
+If we stay with logging I would suggest that we set the cnd log level to `trace`.
+It may be useful to have a few simple scripts in tree that do typical useful grep commands on the logs (unnecessary if we go with tracing).
+
+=== Tracing ===
+
+As mentioned in the issue on GitHub (linked above), logging may not be the correct approach because of the asynchronous nature of cnd.
+The issue is that because logging is event driven and the logs are ordered so it is hard to follow, from reading the log, what events are linked to each other.
+Tracing may be the solution to this but adding tracing requires a non-trivial implementation effort as well as some time investment for all devs on the team to learn how to debug using the trace output and interact with the tracing system.
+On top of this, tracing is most useful for, as the name suggests, 'tracing' a running application.
+Currently, our pain points come from running a failing e2e test, looking at the logs, and not being able to debug the issue.
+If used like this, we will not be leveraging the full potential of tracing, but at least we would be setting the ground work for this style of debugging in the future.
+Having the additional context and structured data may add value and help with debugging.
+
+There are at least two async tracing libraries in the Rust ecosystem, the most popular is by far `tracing` by the tokio folks.
+
+
+== Guidelines/suggestions ==
+
+In summary here are some proposed guidelines to help improve our logs.
+
+I would like to suggest that we attempt to utilise the current logging system fully before deciding to switch to tracing.
+There is an https://github.com/comit-network/comit-rs/pull/1786#issue-355453304[open pull request] to start this, it does a bunch of clean up but can be summarised as follows:
+
+* Do not use targets
+* Set default log level to `Info` for upstream libraries
+* Set default log level to `Trace` for cnd
+* Add a logging trait `Scribe` and use it when logging objects.
+* Audit current log statements, correct log level if required i.e., debug/trace
+
+If/when this pull request merges there will be ongoing work:
+
+
+
+* Add log statements as appropriate to the code base - especially trace statements.
+* Improve `Scribe` implementations as needed.
+
+Perhaps we can run this system for month or two then get team feedback.
--- a/0020-logging.adoc
+++ b/0020-logging.adoc
@@ -47,7 +47,7 @@ Typically one can follow the sequence of log messages to get more context than i
 The solution to this is most likely to use a tracing library instead of logging, see below for more on tracing.
 Currently we typically deal with only a single swap when debugging, once in production we will need to be able to debug with multiple swaps in progress, at a bare minimum this will require the swap_id to be able to be associated with every log message.
 
-Related to context is useful output in the log messages i.e., don't dump byte slices using "{:?}".
+Related to context is useful output in the log messages i.e., don't dump byte slices using `{:?}`.
 
 Suggestions to remedy this (without switching to tracing) include:
 
--- a/0020-logging.adoc
+++ b/0020-logging.adoc
@@ -50,6 +50,7 @@ Currently we typically deal with only a single swap when debugging, once in prod
 Related to context is useful output in the log messages i.e., don't dump byte slices using "{:?}".
 
 Suggestions to remedy this (without switching to tracing) include:
+
 * Implement custom `Debug` for any objects that we would like to print within logs.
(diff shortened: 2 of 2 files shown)


0018-logging.adoc
```diff
@@ -0,0 +1,94 @@
...
+Currently we can control the base log level within the config file.  More fine grained control requires code changes in `cnd/src/logging/initialize.rs`.
+If we stay with logging I would suggest that we set the cnd log level to `trace`.
+It may be useful to have a few simple scripts in tree that do typical useful grep commands on the logs (unnecessary if we go with tracing).
+
+=== Tracing ===
+
+As mentioned in the issue on GitHub (linked above), logging may not be the correct approach because of the asynchronous nature of cnd.
+The issue is that because logging is event driven and the logs are ordered so it is hard to follow, from reading the log, what events are linked to each other.
+Tracing may be the solution to this but adding tracing requires a non-trivial implementation effort as well as some time investment for all devs on the team to learn how to debug using the trace output and interact with the tracing system.
+On top of this, tracing is most useful for, as the name suggests, 'tracing' a running application.
```
thomaseizinger: The most useful aspect about `tracing` is that we can add context to the messages and therefore "build up" the context as the application is running. This means, each layer only needs to add _new information_ to the message because it inherits everything else. Best example for this is the swap id, once set, all the future messages for this one execution path have this swap id. So there is no need to repeat it.

0018-logging.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+Currently we can control the base log level within the config file.  More fine grained control requires code changes in `cnd/src/logging/initialize.rs`.
+If we stay with logging I would suggest that we set the cnd log level to `trace`.
+It may be useful to have a few simple scripts in tree that do typical useful grep commands on the logs (unnecessary if we go with tracing).
+
+=== Tracing ===
+
+As mentioned in the issue on GitHub (linked above), logging may not be the correct approach because of the asynchronous nature of cnd.
+The issue is that because logging is event driven and the logs are ordered so it is hard to follow, from reading the log, what events are linked to each other.
+Tracing may be the solution to this but adding tracing requires a non-trivial implementation effort as well as some time investment for all devs on the team to learn how to debug using the trace output and interact with the tracing system.
```
thomaseizinger: Can you expand on this non-trivial implementation effort?

From reading the documentation, switching to tracing should be a matter of:

- replacing `log::info` with `tracing::info`
- replacing fern with a subscriber like `FmtSubscriber`

0018-logging.adoc
```diff
@@ -0,0 +1,94 @@
...
+Having the additional context and structured data may add value and help with debugging.
+
+There are at least two async tracing libraries in the Rust ecosystem, the most popular is by far `tracing` by the tokio folks.
+
+
+== Guidelines/suggestions ==
+
+In summary here are some proposed guidelines to help improve our logs.
+
+I would like to suggest that we attempt to utilise the current logging system fully before deciding to switch to tracing.
```
thomaseizinger: Without switching to tracing, I will they will literally be useless unless we add the swap id manually to every log statement as soon as there is more than one swap going on.

I would even go as far and say, before we haven't switched to tracing, we are not production ready.

thomaseizinger (dismissed):
Could you please expand on the "non-trivial implementation" effort of switching to tracing?

If we want to invest more time in improving the logs _without_ switching to tracing, I'd like to have a clear statement why we cannot switch. (Mostly so we can fix them and switch :) )

0018-logging.adoc
tcharding: This is an interesting point.  Is the improvement of logging an effort to enable us to debug issues we face _currently_ or is it an effort to be production ready? - if it is to be production ready then I am in total agreeance that we should use tracing.  I assumed we needed a better way to debug issues we are currently facing and were looking for the best and cheapest way to achieve that.  After your response I'll add a section explicitly stating that assumption to the spike.

thomaseizinger: Before you do that, enjoy your holidays :)

0018-logging.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+* error: should only be used when the application really is in trouble. Users are being affected without having a way to work around the issue.
+* warn: should be used when something bad happened, but the application still has the chance to heal itself or the issue can wait a day or two to be fixed.
+* info: should be used to document state changes in the application or some entity within the application.
+* debug: should be used to log any information that helps us identify what went wrong on DEBUG level.
+* trace: log all information that helps us to trace the flow of the application.
+
+=== Useful Context ===
+
+Useful context is not that easy to solve due to the async nature of cnd.
```
D4nte: I would expect the swap id being mentioned here. This is the one information I would like to see on all log lines.

0018-logging.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+Suggestions to remedy this include:
+* Implement custom `Debug` for any objects that we would like to print within logs.
+* Add a `Log` trait for any objects that we wish to log and call the method when logging i.e., `log::info("Got: {}", foo.log())`
+
+=== Controlling Log Levels ===
+
+Currently we can control the base log level within the config file.  More fine grained control requires code changes in `cnd/src/logging/initialize.rs`.
+If we stay with logging I would suggest that we set the cnd log level to `trace`.
+It may be useful to have a few simple scripts in tree that do typical useful grep commands on the logs (unnecessary if we go with tracing).
```
D4nte: I am not convinced about the scripts at this stage.

0018-logging.adoc
```diff
@@ -0,0 +1,94 @@
...
+== Guidelines/suggestions ==
+
+In summary here are some proposed guidelines to help improve our logs.
+
+I would like to suggest that we attempt to utilise the current logging system fully before deciding to switch to tracing.
+There is an https://github.com/comit-network/comit-rs/pull/1786#issue-355453304[open pull request] to start this, it does a bunch of clean up but can be summarised as follows:
+
+* Do not use targets
+* Set default log level to `Info` for upstream libraries
+* Set default log level to `Trace` for cnd
```
D4nte: Users do not need trace, happy to make it by default in e2e tests but I would not want that on a released binary.

0018-logging.adoc
```diff
@@ -0,0 +1,94 @@
...
+* Add a logging trait `Scribe` and use it when logging objects.
```
D4nte: What is the intended gain of such trait?

0018-logging.adoc
tcharding: Fair point.

0018-logging.adoc
tcharding: I thought it might be cleaner than custom implementation of `Display` or `Debug` since these have uses already and we _may_ run into use case conflict.

0018-logging.adoc
tcharding: If you want the swap_id on every log line then I suggest we just drop this whole discussion and go for tracing (where we get context for free) - adding this to every log line will probably lead to
- a macro that wraps the log statements
- passing extra parameters around so we always have access to the swap id when we want to log.

0018-logging.adoc
tcharding: Ok 'non-trivial' might be subjective.  I played with doing this (albeit I did not realise that one had to replace fern) and couldn't get it working so I surmised that it required 'non-trivial' effort.  Will do as you suggest above and re-visit my claim.

0018-logging.adoc
tcharding: Agree 100%, will add this to the spike.

0018-logging.adoc
D4nte: I expect the swap_id to be present at every line yes. Otherwise it means our logging because useless as soon as someone runs 2 swap at the same time.

0018-logging.adoc
tcharding: Here we go, my brain has started to work again - its non-trivial because to get any of the real benefits of tracing we need to add code so that we get context etc.  i.e., we need to design and add spans to the code base.  (spoken as someone who knows very little about tracing.)

0018-logging.adoc
tcharding: Removed.

0018-logging.adoc
tcharding: Removed.

0018-logging.adoc
tcharding: Oh I just mentioned you @D4nte over on the logging PR about this issue ...

tcharding:
This is on ice while I attempt to move cnd over to tracing as a POC.

0020-logging.adoc
```diff
@@ -0,0 +1,95 @@
...
+=== Useful Context ===
+
+Useful context is not that easy to solve due to the async nature of cnd.
+Typically one can follow the sequence of log messages to get more context than is available within the scope of the log statement.
+The solution to this is most likely to use a tracing library instead of logging, see below for more on tracing.
+Currently we typically deal with only a single swap when debugging, once in production we will need to be able to debug with multiple swaps in progress, at a bare minimum this will require the swap_id to be able to be associated with every log message.
+
+Related to context is useful output in the log messages i.e., don't dump byte slices using "{:?}".
+
+Suggestions to remedy this (without switching to tracing) include:
```
da-kami, line 52:
```suggestion
Suggestions to remedy this (without switching to tracing) include:

```
Does not render correctly as list otherwise

0020-logging.adoc
```diff
@@ -0,0 +1,95 @@
...
+* trace: log all information that helps us to trace the flow of the application.
+
+=== Useful Context ===
+
+Useful context is not that easy to solve due to the async nature of cnd.
+Typically one can follow the sequence of log messages to get more context than is available within the scope of the log statement.
+The solution to this is most likely to use a tracing library instead of logging, see below for more on tracing.
+Currently we typically deal with only a single swap when debugging, once in production we will need to be able to debug with multiple swaps in progress, at a bare minimum this will require the swap_id to be able to be associated with every log message.
+
+Related to context is useful output in the log messages i.e., don't dump byte slices using "{:?}".
```
da-kami:
```suggestion
Related to context is useful output in the log messages i.e., don't dump byte slices using `{:?}`.
```

0020-logging.adoc
```diff
@@ -0,0 +1,95 @@
...
+The issue is that because logging is event driven and the logs are ordered so it is hard to follow, from reading the log, what events are linked to each other.
+Tracing may be the solution to this but adding tracing requires a non-trivial implementation effort as well as some time investment for all devs on the team to learn how to debug using the trace output and interact with the tracing system.
+In order to get the benefits of tracing there are code changes that need to be made to define 'spans' and add context.
+On top of this, tracing is most useful for, as the name suggests, 'tracing' a running application.
+Currently, our pain points come from running a failing e2e test, looking at the logs, and not being able to debug the issue.
+If used like this, we will not be leveraging the full potential of tracing, but at least we would be setting the ground work for this style of debugging in the future.
+Having the additional context and structured data may add value and help with debugging.
+
+There are at least two async tracing libraries in the Rust ecosystem, the most popular is by far `tracing` by the tokio folks.
```
da-kami:
```suggestion
There are at least two async tracing libraries in the Rust ecosystem, the most popular is by far https://docs.rs/tracing/0.1.11/tracing/[tracing] by the tokio folks.
```

da-kami (approved):
Pretty straight forward. Some cosmetic change proposals.

0018-logging.adoc
tcharding: Currently working on a POC for this.  I stand by my claim that this is non-trival.

0020-logging.adoc
```diff
@@ -0,0 +1,96 @@
...
+== Guidelines/suggestions ==
+
+In summary here are some proposed guidelines to help improve our logs.
+
+I would like to suggest that we attempt to utilise the current logging system fully before deciding to switch to tracing.
+There is an https://github.com/comit-network/comit-rs/pull/1786#issue-355453304[open pull request] to start this, it does a bunch of clean up but can be summarised as follows:
+
+* Do not use targets
+* Set default log level to `Info` for upstream libraries
+* Set default log level to `Trace` for cnd
```
D4nte, line 85: I think this was already resolved but default should be `info` for cnd, activating `trace` in e2e tests though

D4nte (approved):

luckysori (approved):

thomaseizinger:
@tcharding I am not 100% sure of what the actual outcome is now. The spike recommendation still says to not adopt tracing right away whereas you've also opened a draft PR for tracing now. Please clarify the proposed way forward by creating follow-up tickets of what you think we should do now :)

PR #46, opened 2020-01-02
Add spike for async stack

### LINKED ISSUE #1802: Decide on an async stack
## Problem

We have the choice between `async_std` and `tokio 0.2+` for our async stack.
Historically, `async_std` was ready earlier which is why we use some aspects of it (channels, mutex, etc). The spike should answer the question of which one we want to use. Ideally, we don't have to depend on both of them.

`tokio` is heavily feature flagged, so potentially we _could_ depend on both of them and only bring in what we need.

We still do have to decide on which _Runtime_ we want to use eventually.

## Goal

To make an informed decision for an important aspect like this.

thomaseizinger:
Resolves https://github.com/comit-network/comit-rs/issues/1802.

### DIFF
--- a/0018-async-stack.adoc
+++ b/0018-async-stack.adoc
@@ -1,0 +1,52 @@
+= Decide on an async stack for cnd
+Thomas Eizinger <thomas@coblox.tech>;
+:toc:
+:revdate: 2020-01-02
+
+NOTE: Author: {authors} +
+Date: {revdate} +
+Tracking issue: https://github.com/comit-network/comit-rs/issues/1802[#1802]
+
+== Context
+
+The Rust ecosystem provides several libraries to facilitate an asynchronous programming model.
+The most notable ones are `async-std` and `tokio`.
+
+This spike describes the decision for which one we are going to use in the future.
+Historically (before async/await landed on stable), there was only `tokio` but with 1.39 landing on stable, `async-std` was given to the community.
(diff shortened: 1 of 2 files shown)


tcharding (approved):
You haven't touched on the fact that we cannot, AFAICT, depend on _both_ tokio 1 and 2 at the same time.  And we have to depend on tokio 1 while ever we use libraries that depend on tokio 1 i.e., libp2p.  If I'm not mistaken this is why we have the `async-std` stuff we currently have.  We are at the mercy of our dependencies are we not?  I do however agree with this spike that we should use tokio 2 only at some stage.

thomaseizinger:
> You haven't touched on the fact that we cannot, AFAICT, depend on _both_ tokio 1 and 2 at the same time. And we have to depend on tokio 1 while ever we use libraries that depend on tokio 1 i.e., libp2p. If I'm not mistaken this is why we have the `async-std` stuff we currently have. We are at the mercy of our dependencies are we not? I do however agree with this spike that we should use tokio 2 only at some stage.

`tokio-compat` was recently introduced which provides an executor that can handle tokio 0.1 and tokio 0.2 futures :)

See here: https://github.com/comit-network/comit-rs/pull/1804

luckysori (approved):

D4nte (approved):

PR #47, opened 2020-01-20
introducing an easy-to-read naming scheme for our protocols

da-kami:
We have another HALight, ahm Highlight for you...

And yes, we are aware that the ticket said we should not create a spike. But creating an RFC right away seems a bit too fast 😉

### DIFF
--- a/0021-protocol-naming.adoc
+++ b/0021-protocol-naming.adoc
@@ -1,5 +1,5 @@
 = Introducing a protocol naming scheme
-Thomas Eizinger <thomas@coblox.tech>; Daniel Karzel<daniel.karzel@coblox.tech>
+Thomas Eizinger <thomas@coblox.tech>; Daniel Karzel <daniel.karzel@coblox.tech>;
 :toc:
 :revdate: 2020-01-20
 
--- a/0021-protocol-naming.adoc
+++ b/0021-protocol-naming.adoc
@@ -11,10 +11,10 @@ Tracking issue: https://github.com/comit-network/RFCs/issues/130[#130]
 
 We propose the following protocol names for Alpha and Beta ledger:
 
-- `HAN` - HTCL for Asset on Native
-- `HErc20` - HTLC for ERC20
-- `HOmni` - HTLC for OMNI
-- `HALight` - HTLC for Lightning
+- `HAN` - (atomic swaps using ) **H**TLCs for **A**ssets that are **N**ative to the ledger
+- `HErc20` - (atomic swaps using ) **H**TLCs for the **Erc20** asset
+- `HOmni` - (atomic swaps using ) **H**TLCs for the **Omni** asset
+- `HALight` - (atomic swaps using ) **H**TLCs for **A**ssets on the **Light**ning ledger
 
 A swap would then be defined by a combination of two protocols, here are three examples:
 
--- a/0021-protocol-naming.adoc
+++ b/0021-protocol-naming.adoc
@@ -1,0 +1,46 @@
+= Introducing a protocol naming scheme
+Thomas Eizinger <thomas@coblox.tech>; Daniel Karzel<daniel.karzel@coblox.tech>
+:toc:
+:revdate: 2020-01-20
+
+NOTE: Author: {authors} +
+Date: {revdate} +
+Tracking issue: https://github.com/comit-network/RFCs/issues/130[#130]
+
+== Proposal
+
+We propose the following protocol names for Alpha and Beta ledger:
+
+- `HAN` - HTCL for Asset on Native
+- `HErc20` - HTLC for ERC20
+- `HOmni` - HTLC for OMNI
+- `HALigth` - HTLC for Lightning
+
+A swap would then be defined by a combination of two protocols, here are three examples:
+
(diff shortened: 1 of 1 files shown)


0021-protocol-naming.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+NOTE: Author: {authors} +
+Date: {revdate} +
+Tracking issue: https://github.com/comit-network/RFCs/issues/130[#130]
+
+== Proposal
+
+We propose the following protocol names for Alpha and Beta ledger:
+
+- `HAN` - HTCL for Asset on Native
```
tcharding: Perhaps, 'HTLC Native Asset' - HNA

0021-protocol-naming.adoc
```diff
@@ -0,0 +1,46 @@
...
+- `HErc20` - HTLC for ERC20
+- `HOmni` - HTLC for OMNI
+- `HALigth` - HTLC for Lightning
```
tcharding: Is this meant to be 'HTLC for Asset on Lighting'?  Perhaps 'HAL'?

0021-protocol-naming.adoc
tcharding: The tokens would become 'HTLC Custom Asset' - HCA

D4nte: Star Wars theme now?

D4nte: I agree with @tcharding the `HTLC for Asset on Native` does not sound English.
`HTLC for Native Asset` sounds fine.

0021-protocol-naming.adoc
da-kami: we had `HAL` in the beginning, but that is limiting if you assume that another special implementation for a ledger starting with `L` could exist. For specific implementation names we opted for longer names.

0021-protocol-naming.adoc
da-kami: `HAN` just reads and sounds nicer, that's why we changed the order. In the beginning we had `NAH` (which is like a fun version of "no" in Austrian dialect). HAN just makes it easy to pronounce.

0021-protocol-naming.adoc
da-kami:
```suggestion
- `HALight` - HTLC for Lightning
```

0021-protocol-naming.adoc
```diff
@@ -0,0 +1,46 @@
+= Introducing a protocol naming scheme
+Thomas Eizinger <thomas@coblox.tech>; Daniel Karzel<daniel.karzel@coblox.tech>
```
thomaseizinger:
```suggestion
Thomas Eizinger <thomas@coblox.tech>; Daniel Karzel <daniel.karzel@coblox.tech>;
```

thomaseizinger (approved):
Awesome :)

0021-protocol-naming.adoc
D4nte: Nonetheless "HTLC for asset on native" sounds weird.

thomaseizinger: Agree with @da-kami here. The abbreviation is massively more important than how you expand it. The expansion of `HAN` could also read:

(Atomic swap using) **H**tlcs of **A**ssets that are **N**ative to the ledger

HAN can be easily pronounced, is short and contains all the important information to differentiate it from other protocols.

At this stage, I would suggest for reviewers to make a counter proposal that is at least equally good (measured by all the aspects mentioned in the spike) if they are not happy with the result.

0021-protocol-naming.adoc
tcharding: Reasonable point.

thomaseizinger:
@da-kami I pushed a commit that slightly revises the expansion of the abbreviations.

D4nte (commented):
I have reviewed and provided my input.

luckysori (commented):
Open question: what is the word asset (and the letter A) adding if everything that we deal with _is_ an asset?

thomaseizinger:
@luckysori That is a good question! First thoughts I have on these are:
We could at some point have protocols that don't deal with assets? Opening channels maybe?

It doesn't hurt and makes the acronym sound better 🤷‍♂️

@d4nte, are you happier with the update of the expanded text?

D4nte (approved):
LGreatTM

0021-protocol-naming.adoc
tcharding: > HAN just reads and sounds nicer, that's why we changed the order.
Don't tell fibs, this is a reference to the Chinese Han dynasty, you slipped it in hoping we wouldn't notice :)

tcharding: I agree with @D4nte that "HTLC for asset on native" sounds rubbish but also with the others that 'HAN' is a better acronym.  "HTLC Asset Native" is fine though and satisfies both camps :)

D4nte: You are late to the party they already changed the proposal ;)

tcharding:
@da-kami MERGE yo!

PR #51, opened 2020-02-01
A new communication protocol for COMIT

### LINKED ISSUE #1827: Design a new communication protocol for COMIT
## Problem

The current communication protocol was our first attempt. Since then, we learned many things and also adopted libp2p, which made several aspects of the design redundant.

### Be "libp2p-native":

- leverage the "protocol" concept of libp2p
- follow patterns such as oneshot substreams where applicable

### Leverage the negotiation phase

We've discovered that very likely, there will always be a negotiation phase before COMIT even gets to know about the swap.
The new communication protocol should leverage this fact to simplify the communication by avoiding (re-)transmission of data the applications already know (the famous, auto-accept problem).
To solve this, the spike will need to identify, which data definitely needs to be shared among the nodes and is not in control by the applications.
We've already identified that at the very least, this includes:

- secret-hash
- transient identities (Bitcoin for now, proposal for Ethereum is available)

### Modularity & open-closed principle

At the moment, the communication protocol is a single request-response handshake.
This imposes a lot of complexity on the message format, as it has to handle all kinds of different scenarios.
Ideally, the new communication protocol has a more modular approach to allow for different messages to be sent, depending on the ledgers, assets and swap-protocols being involved.
In a way, this relates to the open-closed principle.
Extensions to the communication protocol should happen in the form of new messages instead of modifying existing ones.

### Support for more ledgers

While designing the new protocol, one should take into account the needs for support Lightning and Grin.

thomaseizinger:
Resolves https://github.com/comit-network/comit-rs/issues/1827.

### DIFF
--- a/0023-communication-protocol.adoc
+++ b/0023-communication-protocol.adoc
@@ -148,9 +148,7 @@ For HTLC-based protocols like HAN, HErc20 and HALight, Alice generates a secret.
 The communication protocol must provide a way for Alice to share the hash of this secret with Bob.
 We also have to take into account that in the future, we will have swap protocols that depend on other cryptographic elements like signatures.
 
-==== Sharing identities
-
-Atomic swap protocols center around the idea of providing a time window in which the change in ownership of assets is atomic.
+Atomic swap protocols in general center around the idea of providing a time window in which the change in ownership of assets is atomic.
 A core aspect of these protocols are identities.
 The communication protocol must provide a way for both parties to share identities with the other party.
 For on-chain (i.e. Layer 1) protocols like HAN and HErc20, this is always a pair of identities (redeem + refund).
--- a/0023-communication-protocol.adoc
+++ b/0023-communication-protocol.adoc
@@ -2,6 +2,8 @@
 Thomas Eizinger <thomas@coblox.tech>;
 :toc:
 :revdate: 2020-01-28
+:sectnums:
+:sectnumlevels: 5
 
 NOTE: Author: {authors} +
 Date: {revdate} +
@@ -340,5 +342,33 @@ Once each party sent a finalize message AND received one, implementations should
 
 == Discussion points
 
-- Should we use multihash for the swap digest?
-- how to compute the swap digest?
+This section includes discussion topics that remain to be answered.
+
+=== Should we use multihash for the swap digest
+
+<<Announce-message,Section 3.2>> introduces the concept of a swap digest.
+In order to be forwards compatible with different hash algorithms, it proposes to encode the merkle root as a multihash.
+That way, the hash is self-describing and the receiving party knows, which hash algorithm needs to be used to compute the digest.
+
+Do we need the flexibility of changing the hash function that is used to compute the swap digest without a breaking change to the protocol?
+We could always just create `/comit/swap/announce/2.0.0` that defines a different hash function.
+
+=== How to compute the swap digest
+
+It is still to be defined, how exactly the swap digest will be computed.
+In particular, we need to define:
+
+* the order in which the tree of information about the swap is arranged
+* which parts of the swap are included in the merkle tree
+
+Especially the last point is important to discuss.
+Consider the following:
+
+Given the above protocols, it is should be possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
+Considering our current implementation, we would still generate transient bitcoin identities to use in the HTLC and only pay to the user's wallet upon redeeming.
+Both of these requirements combined raises the question:
+how does cnd know, which of the parameters submitted via the API are "negotiated" (i.e. also known to the other party) and which ones where added locally by the user?
+Only the "negotiated" ones must be added to the swap digest, otherwise the digests won't match.
+
+The only answer I have for this so far is that we should just always exlude identities from the swap digest.
+Thinking about it, why would any negotiation protocol ever include identities?
--- a/0023-communication-protocol.adoc
+++ b/0023-communication-protocol.adoc
@@ -359,13 +359,9 @@ We could always just create `/comit/swap/announce/2.0.0` that defines a differen
 === How to compute the swap digest?
 
 It is still to be defined, how exactly the swap digest will be computed.
-In particular, we need to define:
+In particular, we need to define which parts of the swap are included in the hash list.
 
-* the order in which the tree of information about the swap is arranged
-* which parts of the swap are included in the hash list
-
-Especially the last point is important to discuss.
-Consider the following:
+==== Identities
 
 Given the above protocols, it is possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
 Considering our current implementation, we would still generate transient bitcoin identities to use in the HTLC and only pay to the user's wallet upon redeeming.
@@ -375,3 +371,19 @@ Only the "negotiated" ones must be added to the swap digest, otherwise the diges
 
 The only answer I have for this so far is that we should just always exlude identities from the swap digest.
 Thinking about it, why would any negotiation protocol ever include identities?
+
+==== Collision resistance
+
+The digest should be as collision resistant as possible.
+I would propose to include:
+
+- alpha-ledger
+- beta-ledger
+- alpha-asset
+- beta-asset
+- alpha-protocol
+- beta-protocol
+- (depending on the protocol) alpha-expiry
+- (depending on the protocol) beta-expiry
+- alice-identity (libp2p ID)
+- bob-identity (libp2p ID)
--- a/0023-communication-protocol.adoc
+++ b/0023-communication-protocol.adoc
@@ -134,24 +134,148 @@ Opening substreams is very cheap compared to network connections.
 While there is still some inefficiency in terms of opening substreams, a more flexible design seems to be desirable over some added latency.
 Plus, https://github.com/libp2p/specs/tree/master/identify[work is underway] to make this more efficient.
 
-=== Identifying the moving parts
+=== Communication requirements
 
-Marry one protocol per swap with oneshot pattern
+In order to design the ideal communication protocol, we have to identify which data we want to transfer and why.
+This section aims to achieve that.
 
-=== Agreeing on a common identifier
+==== Verifying the details of the swap
 
-- make sure we talk about the same thing (swap digest)
-- who proposed the identifier?
+One of the primary requirements for the communication protocol is for both parties to verify that they talk about the same swap.
+This is crucial in avoiding unnecessary transactions, as long as both parties are willing to actually swap.
 
-=== How we know we are ready to start the swap?
+==== Sending over protocol-specific data
 
-- finalize message
+For HTLC-based protocols like HAN, HErc20 and HALight, Alice generates a secret.
+The communication protocol must provide a way for Alice to share the hash of this secret with Bob.
+We also have to take into account that in the future, we will have swap protocols that depend on other cryptographic elements like signatures.
 
-=== Error handling
+==== Sharing identities
 
-- timeouts
-- close on announce substream
+Atomic swap protocols center around the idea of providing a time window in which the change in ownership of assets is atomic.
+A core aspect of these protocols are identities.
+The communication protocol must provide a way for both parties to share identities with the other party.
+For on-chain (i.e. Layer 1) protocols like HAN and HErc20, this is always a pair of identities (redeem + refund).
+For Layer 2 protocols like HALight, the identity is a single public key: the identifier of the user's lnd instance.
+We have to accommodate for these differences in protocols.
 
 == Recommendation
 
-[Based on the research, try to make a recommendation for one of the proposed solution. If you can't, call in a meeting to decide on an outcome]
+Based on the research above, this section presents a proposal for a set of libp2p protocols that can be combined to offer the required functionality.
+
+=== Announcing a swap
+
+Protocol-identifier: `/comit/swap/announce/1.0.0`
+
+This protocol includes two messages:
+
+. The announce message
+. The confirmation message
+
+==== Announce-message
+
+[source,json]
+----
+{
+  "swap_digest": "12202c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae"
+}
+----
+
+This message is always sent by Alice.
+See <<Assumptions and goals,assumptions point 3>> for further explanation.
+
+The `swap_digest` field is a https://github.com/multiformats/multihash[multihash-encoded] merle root that contains all details of the swap.
+It remains to be defined, how the digest is exactly computed but the idea is the following:
+
+. Take all of the swaps information _known to the COMIT node_ (alpha ledger, alpha asset, protocols, etc) and arrange it as a tree.
+. Compute the merkle root by hashing all of the nodes together
+
+The idea is that, independently of what information about the swap is given to the COMIT node (imagine different usecases, with different negotiation protocols on top of COMIT), there is a deterministic way of computing this digest.
+If both parties, independently of each other, compute the same digest, they must be knowledgeable of the same parameters.
+This should give us enough certainty to start executing the swap assuming the other party is in good will.footnote:[It is pretty much impossible to fully protected against bad will upfront.]
+Only sending a digest has several advantages over sending the actual data of the swap for comparison:
+
+- The message format is much simpler and can be reused for every swap-protocol, independent of the cryptographic protocol in use.
+- The message format is agnostic over what the nodes are actually committing to.
+
+The nodes know, which data they used to compute the digest.
+Consequently, they are also aware of what is still missing to start the swap.
+Imagine a negotiation protocol that doesn't include expiriesfootnote:[Despite the fact that they are price-relevant.].
+The nodes could:
+
+.. either default to certain expiries and include them in the digest
+.. or simply not include them in the digest.
+
+In case of a), a matching digest means both nodes used the same default expiries.
+In case of b), a matching digest means both nodes are aware that the expiries are still missing and they will have to somehow negotiate them.
+
+==== Confirmation message
+
+[source,json]
+----
+{
+  "swap_id": "713b0c93-53a1-46b2-9418-de8d46750b79"
+}
+----
+
+In case the digest matches, Bob generates a UUIDv4 and sends it back to Alice.
+If the digest does not match, Bob closes the substream without sending any message.
+
+You may wonder, why Bob generates an ID on top of the digest that Alice shared with him.
+The reason is simple: while highly unlikely, the digest is not unique.
+
+Having Bob generating a random, dedicated identifier for this swap allows us:
+
+.. Make it explicit that Bob agrees to this swap and is knowledgeable of the same parameters as Alice.
+.. Establish a dependency between this libp2p protocol and all others.
+The other libp2p protocols use the swap-id generated by Bob to identify the swap.
+None of the parties can construct a meaningful message without going through this handshake, making it clear that this handshake has to happen before everything else.
+
+=== Sharing the secret hash
+
+Protocol-identifier: `/comit/swap/secrethash/1.0.0`
+
+This protocol is only to be used by Alice, as she is the one generating the secret.
+
+This protocol only includes a single message to share the secret hash.
+If you paid close attention to the protocol identifier, you may have realized that it is actually protocol agnostic:
+It neither mentions HAN, HErc20 or any of the others.
+The reason is simple: We don't need different messages for different protocol combinations if they always just need to exchange a secret hash.
+Using this libp2p protocol implies that the referenced swap is HTLC-based.
+
+==== Secret hash message
+
+[source,json]
+----
+{
+  "swap_id": "713b0c93-53a1-46b2-9418-de8d46750b79",
+  "secret_hash": "122003f2fb5188982860e6bdaa4cebdae5509311da248c4bd7ba1126f78a881cb816"
+}
+----
+
+The `swap_id` refers to the swap-id generated by Bob in the confirmation message.
+The `secret_hash` is a https://github.com/multiformats/multihash[multihash] of the secret generated by Alice.
+
+We use a multi-hash to allow re-use of this message for any HTLC-based protocol, independent of the hash algorithm that is used.
+It is important to note that for a given swap protocol pair, only a single hash algorithm will be valid.
+The above example uses SHA256, which is the hash algorithm we currently use for HAN, HErc20 and HALight.
+
+Implementations MUST verify that the hash algorithm used to encode this multihash is the same one as the one used in the swap protocol.
+Implementations MUST also take care to not being tricked into using a weaker hash algorithm.
+However, this is very unlikely anyway because the hash algorithm used by HTLCs is usually not configurable.
+
+=== Sharing identities
+
+Which identities need to be shared depends on:
+
+* the protocols of the swap
+* the alpha and beta ledger
+* the role
+
+`/comit/bitcoin/identity/1.0.0`
+`/comit/ethereum/identity/1.0.0`
+`/comit/lightning/identity/1.0.0`
+
+== Discussion points
+
+- Should we use multihash for the swap digest?
--- a/0023-communication-protocol.adoc
+++ b/0023-communication-protocol.adoc
@@ -373,7 +373,7 @@ In particular, we need to define:
 Especially the last point is important to discuss.
 Consider the following:
 
-Given the above protocols, it is should be possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
+Given the above protocols, it is possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
 Considering our current implementation, we would still generate transient bitcoin identities to use in the HTLC and only pay to the user's wallet upon redeeming.
 Both of these requirements combined raises the question:
 how does cnd know, which of the parameters submitted via the API are "negotiated" (i.e. also known to the other party) and which ones where added locally by the user?
--- a/0023-communication-protocol.adoc
+++ b/0023-communication-protocol.adoc
@@ -134,7 +134,7 @@ In other words, no need to have `type` fields inside the message, the protocol a
 * cheap:
 Opening substreams is very cheap compared to network connections.
 While there is still some inefficiency in terms of opening substreams, a more flexible design seems to be desirable over some added latency.
-Plus, https://github.com/libp2p/specs/tree/master/identify[work is underway] to make this more efficient.
+Plus, https://github.com/libp2p/specs/pull/205/files[work is underway] to make this more efficient.
 
 === Communication requirements
 
--- a/0023-communication-protocol.adoc
+++ b/0023-communication-protocol.adoc
@@ -218,10 +218,10 @@ In case of b), a matching digest means both nodes are aware that the expiries ar
 }
 ----
 
-In case the digest matches, Bob generates a UUIDv4 and sends it back to Alice.
-If the digest does not match, Bob closes the substream without sending any message.
+In case Bob knows about a swap with the same digest, he generates a UUIDv4 and sends it back to Alice.
+If he doesn't know about a swap with this digest, he closes the substream without sending any message.
 
-You may wonder, why Bob generates an ID on top of the digest that Alice shared with him.
+You may wonder why Bob generates an ID on top of the digest that Alice shared with him.
 The reason is simple: while highly unlikely, the digest is not unique.
 
 Having Bob generating a random, dedicated identifier for this swap allows us:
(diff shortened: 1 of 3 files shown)


0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+With the https://github.com/comit-network/spikes/blob/master/0021-protocol-naming.adoc[protocol naming spike], we've made it more explicit that COMIT does not support one but several cryptographic protocols at the moment.
+In particular, we currently support:
+
+* HAN for Bitcoin
+* HAN for Ethereum
+* HErc20 for Ethereum
+
+We've managed to identify a protocol (HAN) that can be applied to both of the ledgers that we currently support.
+Even though this allows for some reuse, when you take a look at the actual interactions with the blockchain, HAN for Bitcoin and HAN for Ethereum differ (you cannot broadcast a Bitcoin transaction on Ethereum and vice versa).
+Given that there is always two protocols that need to be combined for an atomic swap, our total number of supported protocols (considering the full swap as a black box) is 6 (2 * 3).footnote:[We don't allow HAN(Ethereum)-HErc20 at the moment, so this number drops to 4 in reality.]
```
tcharding, line 82: I don't understand the math here?  Is this the combinations of protocols that we support?  If so shouldn't it by 4, since we only support cross-chain swaps?

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+A communication protocol within COMIT defines how two COMIT nodes interact with each other.
+Those (libp2p-based) protocols facilitate the exchange of information for the purpose of then following a cryptographic protocol.
+
+These two kinds of protocols have a relationship to each other.
+In the past we went with a N-to-1 relationship: we designed a single communication protocol that was supposed to handle communication for all kinds of cryptographic protocols.
```
tcharding: May read nicer using 1-to-N instead of N-to-1 since that is the order of the English description that follows.

0023-communication-protocol.adoc
tcharding: Oh, this may be a typo.  You say 4 below.

tcharding (approved):
Solid work @thomaseizinger.  As per current discussion on Basecamp I treated this as a 50% review and did not touch typos/grammar or anything too trivial.  You may like to fix up the typo `s/merle/merkle/` though :)  Really thorough, was a pleasure to read.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+Even though we sweated blood in the design of the original communication protocol, we quickly learned that it is not as extensible and flexible as we thought it would to be.
+
+=== What is the ideal relationship between cryptographic and libp2p protocols?
+
+The main problem with the design of the original communication protocol was that we tried to anticipate the future a bit too much.
+Even thought it is important to allow for future extensions, one has to be careful about how the extension mechanism works.
+
+Let's consider the following requirement:
+
+. In order to execute a certain cryptographic protocol, COMIT nodes may want to exchange information.
```
D4nte: I am not sure I understand your point here.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+These protocols are small and focused, avoiding the trap of over-engineering them.
+* simpler client code:
+If there is only one possible message that can be sent, its shape can be hardcoded by the client.
+Software that uses libp2p has to cater for several protocols being active concurrently anyway.
+Using the protocol layer as the level with which messages vary makes many things a lot simpler.
+In other words, no need to have `type` fields inside the message, the protocol already identifies the type.
+* cheap:
+Opening substreams is very cheap compared to network connections.
+While there is still some inefficiency in terms of opening substreams, a more flexible design seems to be desirable over some added latency.
+Plus, https://github.com/libp2p/specs/tree/master/identify[work is underway] to make this more efficient.
```
D4nte: wrong link?

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+One of the primary requirements for the communication protocol is for both parties to verify that they talk about the same swap.
+This is crucial in avoiding unnecessary transactions, as long as both parties are willing to actually swap.
+
+==== Sending over protocol-specific data
+
+For HTLC-based protocols like HAN, HErc20 and HALight, Alice generates a secret.
+The communication protocol must provide a way for Alice to share the hash of this secret with Bob.
+We also have to take into account that in the future, we will have swap protocols that depend on other cryptographic elements like signatures.
+
+==== Sharing identities
```
D4nte: I am not sure to understand the difference with the previous section as identity are protocol-specific data.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+For the HALight protocol, the parties only need to exchange one identity: their LND pubkey.
+
+We can define very simple protocols to allow for this information to be exchanged.
+These are the protocols we need:
+
+==== Protocols for sharing identities
+
+Protocol identifier: `/comit/swap/identity/bitcoin/1.0.0`
+
+[source,json]
```
D4nte, line 278: Nitpick: use Javascript to allow comments.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+This section includes discussion topics that remain to be answered.
+
+=== Should we use multihash for the swap digest
+
+<<Announce-message,Section 3.2>> introduces the concept of a swap digest.
+In order to be forwards compatible with different hash algorithms, it proposes to encode the merkle root as a multihash.
+That way, the hash is self-describing and the receiving party knows, which hash algorithm needs to be used to compute the digest.
+
+Do we need the flexibility of changing the hash function that is used to compute the swap digest without a breaking change to the protocol?
+We could always just create `/comit/swap/announce/2.0.0` that defines a different hash function.
```
D4nte: It is a difficult question. On one hand I believe that the hash algorithm should be treated like the expiries. It can be "defaulted" and hence not present in the merkle tree or always present with the same value.

By using a multihash we are taking a characteristic of the swap and choose/specify it in a separate place than the rest of the characteristics (amounts, expiries).

Moreover, it means that we need to have added error handling logic there. I think it should all be in the Merkle tree.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+- The message format is much simpler and can be reused for every swap-protocol, independent of the cryptographic protocol in use.
+- The message format is agnostic over what the nodes are actually committing to.
+
+The nodes know, which data they used to compute the digest.
+Consequently, they are also aware of what is still missing to start the swap.
+Imagine a negotiation protocol that doesn't include expiriesfootnote:[Despite the fact that they are price-relevant.].
+The nodes could:
+
+.. either default to certain expiries and include them in the digest
+.. or simply not include them in the digest.
```
D4nte, line 206: I think this is a dangerous root as it means we cannot change the default once they are fixed (as they are not part of the merkle root, they cannot be changed).

If the default as per this communication protocol become obsolete then we would end up having to advise to always specify expiries and go against such default (as they cannot be changed).

I'd recommend to include them in the digest.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+Consider the following:
+
+Given the above protocols, it is should be possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
+Considering our current implementation, we would still generate transient bitcoin identities to use in the HTLC and only pay to the user's wallet upon redeeming.
+Both of these requirements combined raises the question:
+how does cnd know, which of the parameters submitted via the API are "negotiated" (i.e. also known to the other party) and which ones where added locally by the user?
+Only the "negotiated" ones must be added to the swap digest, otherwise the digests won't match.
+
+The only answer I have for this so far is that we should just always exlude identities from the swap digest.
+Thinking about it, why would any negotiation protocol ever include identities?
```
D4nte, line 373: I am in favour of excluding identities from the digest and keep the `/comit/swap/identity` protocol defined above as unique way to communicate identities.

Either you define the `/comit/swap/identity` protocols and have identity communicated at negotiation time and set in the merkle digest, or you do not have them in the digest and exchange them via the `/comit/swap/identity` protocol.

It would unnecessary complex to have both (part of merkle root AND exchangeable via `/comit/swap/identity`.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+=== Is there a better algorithm than a merkle tree for the swap digest?
+
+A merkle tree was mainly chosen because the size of the merkle root is constant, regardless of how big the tree was.
+A downside of the merkle tree is that it only yields the same digest if the tree is arranged in exactly the same way.
+This leaves room for error in implementations and specifications.
```
D4nte: Indeed it does, however there is no security risk to this, only downside would be 2 incompatible clients that cannot execute swaps with each others.

As we already know that it is an important piece of information then we should ensure that it is part of the RFC and kept up to date.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+
+Is there an algorithm that can fulfill the same role but is independent of the ordering of its parameters?
```
D4nte: One way would be to still use the merkle tree but use a generic algorithm to order the parameters that can be applied for present and future COMIT protocols.

D4nte (commented):
LGTM, I gave much opinion even if it's a draft 🙃

0023-communication-protocol.adoc
da-kami: For clarification:

He means these 4:

```
HAN(Bitcoin)-HAN(Ether)
HAN(Ether)-HAN(Bitcoin)
HErc20-HAN(Bitcoin)
HAN(Bitcoin)-HErc20
```

+2 theoretical ones that do not make sense because you can just do that on Ethereum without needing atomic swaps:

```
HErc20-HAN(Ether)
HAN(Ether)-HErc20
```

0023-communication-protocol.adoc
da-kami: Let me try to add some information, let me know if it helps:

Cryptographic protocols may require information exchange such as e.g. the secret-hash and certain identities for HAN. But the information exchanged may differ between different cryptographic protocols. 

Thus,  the question is, how to design the **communication protocol** given different information exchange in cryptographic protocols? Thomas talks about 2 extremes, we can create the communication protocol very generic so one can pack different information into one message of one protocol, or we can split it into several protocols that have a specific concern and only one message.

0023-communication-protocol.adoc
da-kami: I agree this section needs some explanation why it is separate.

From a discussion I had with @thomaseizinger before he left. Yes, in the context of atomic swaps identities are protocol-specific parameters. But the identities are actually independent of the swap. They exist on the ledger and are not purely created for the swap (like e.g. the secret). Thus we could handle them separate. A protocol to exchange e.g. a Bitcoin identity could be used for all kinds of use cases, not only atomic swaps.

da-kami (commented):
still in the middle of reviewing, but submitting current status.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+* which parts of the swap are included in the merkle tree
+
+Especially the last point is important to discuss.
+Consider the following:
+
+Given the above protocols, it is should be possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
+Considering our current implementation, we would still generate transient bitcoin identities to use in the HTLC and only pay to the user's wallet upon redeeming.
+Both of these requirements combined raises the question:
+how does cnd know, which of the parameters submitted via the API are "negotiated" (i.e. also known to the other party) and which ones where added locally by the user?
+Only the "negotiated" ones must be added to the swap digest, otherwise the digests won't match.
```
da-kami, line 370: I like the proposed `announce` protocol for its simplicity.
Two comments, just for the sake of awareness: 

1. We have to very carefully select which parameters are part of the digest to avoid **collisions** (i.e. same ledgers+assets+amounts+protocol, but different identities); for a high-frequency trader it is likely to have a lot of swaps in progress. Note that this issue is a general issue that is independent of calculation of the digest, but has to be taken into account when actually designing the digest.

I assume that a (unlikely) collision (i.e. two swaps with the same digest being in the pre-announce state at the same time) is in the responsibility of the application on top of COMIT?

2. There is no way to "give feedback" to the other party any more. When sending the swap parameters we see what the other party sends and can provide feedback on what is missing/wrong. When sending a computed digest we only find a match or we fail. There is no feedback whatsoever. We can see this as advantage (more resilient to spam) and disadvantage (no feedback to a different COMIT implementation).

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+=== Announcing a swap
+
+Protocol-identifier: `/comit/swap/announce/1.0.0`
+
+This protocol includes two messages:
+
+. The announce message
+. The confirmation message
+
+==== Announce-message
```
da-kami, line 171: Note: I moved my comments concerning the `announce` protocol to the Discussion Points section because they seemed to fit better there.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,388 @@
...
+=== Should we use multihash for the swap digest
+
+<<Announce-message,Section 3.2>> introduces the concept of a swap digest.
+In order to be forwards compatible with different hash algorithms, it proposes to encode the merkle root as a multihash.
+That way, the hash is self-describing and the receiving party knows, which hash algorithm needs to be used to compute the digest.
+
+Do we need the flexibility of changing the hash function that is used to compute the swap digest without a breaking change to the protocol?
+We could always just create `/comit/swap/announce/2.0.0` that defines a different hash function.
+
+=== How to compute the swap digest?
```
da-kami, line 359: The definition of how to calculate the digest will be per protocol-combination I suppose? I.e. one definition for HAN-HAN, one for HErc20-HAN, (...). Correct me if I got this wrong.

da-kami (approved):
Great work @thomaseizinger !

I added a few comments and raised questions in the "How to compute the swap digest?" section. Non-blocking.

0023-communication-protocol.adoc
D4nte: I think my issue is more with the wording: "Let's consider the following requirement" invite to consider an example to make the explanation clearer to the reader.

Upon reading this section, I am more confused then if the example were not here.

0023-communication-protocol.adoc
D4nte: Let's add more info there then please.

0023-communication-protocol.adoc
thomaseizinger: What this statement aimed to make clearer is that the communication protocol is _motivated_ by us wanting to execute cryptographic protocols. Maybe this is so obvious that it doesn't need mentioning?

0023-communication-protocol.adoc
thomaseizinger: I am using comments, I am using asciidoc syntax. Reader the rendered version :)

0023-communication-protocol.adoc
thomaseizinger: I am talking about the swap digest here, not the hash algorithm of the HTLC.

0023-communication-protocol.adoc
thomaseizinger: Did you read the whole paragraph? :sweat_smile: 

This section is meant to serve as an example of how the digest works and is not a proposal of whether the expiries should be included.
Actually, I specifically state that they should be included just 3 lines above our comment it says:

> Imagine a negotiation protocol that doesn't include expiries.

With a footnote:

> Despite the fact that they are price-relevant.

0023-communication-protocol.adoc
thomaseizinger: I think I agree with you on the outcome of not including identities in the digest. My reasons are different though:

If the digest where to include identities, we assume that every negotiation protocol includes them. More generally: everything that goes into the digest is assumed to be negotiated (or to have defaults in the implementation that are the same in all implementations, f.e. because they are defined in the COMIT spec)

Anyway, I think we agree that the identities should not be part of the negotiation. They conceptually belong to the execution phase.

0023-communication-protocol.adoc
thomaseizinger: Solved that one by using a hash list.

0023-communication-protocol.adoc
thomaseizinger: Solved that using a hash list instead.

0023-communication-protocol.adoc
thomaseizinger: Regarding 1)

Yes, the swap digest is not an identifier. Hence, any application using this communication protocol has to come with its own means of uniquely identifying data that flows through the system. For cnd, this likely means assigning a local ID to any request made on the REST API.

Any party receiving such an announcement message with the same digest as a previous one MUST treat them as different swaps and assign a new ID.

Regarding 2)

I would suggest to add messages for feedback later on, once we have clear needs to avoid a similar mess that we currently have with the unused decline reason.

0023-communication-protocol.adoc
thomaseizinger: The elements of the hash list partly depend on the protocols in use. I would suggest to define the serialization of the individual elements separately and then define per protocol, what needs to be included:

- alpha-leger
- beta-ledger
- alpha-asset
- beta-asset
- alpha-protocol
- beta-protocol
- alpha-expiry
- beta-expiry
...

Since it is a flat list, we can put whatever we want in there :)

thomaseizinger:
@D4nte @tcharding @da-kami I've re-requested a review but also force pushed to resolve the merge conflicts. Only the last three commits are actually new - the rest is unchanged :)

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+
+While designing the new protocol, one should take into account the needs for support Lightning and Grin.
+
+== Research
+
+=== Assumptions and goals
+
+I've started the research and design of the new communication protocol with the following assumptions and goals in mind.
+
+. Avoid interaction as part of the communication protocol.
```
bonomat: Do you mean _interactions_ ala request+response? 
Do you assume there will be an `ACK` message back or is it fire and forget (maybe libp2p offers something here already)?

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+. In order to execute a certain cryptographic protocol, COMIT nodes may want to exchange information.
+
+The important thing to realize here is that not every cryptographic protocol necessarily requires communication.
+Almost certainly, we should not design a single communication protocol that caters for all possible cryptographic protocols.
+
+=== What is the scope of a single communication protocol?
+
+Knowing that we likely want more than one communication protocol, the question is, what is the scope of one of those libp2p protocols?
+
+In order to not fall into the same trap as with the original communication protocol, we can just opt to not design libp2p protocols for things that we don't (yet) know.
```
bonomat: Do you mean this? 
```suggestion
In order to not fall into the same trap as with the original communication protocol, we can just opt to design libp2p protocols for things that we don't (yet) know.
```

D4nte (approved):

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+----
+
+This message is always sent by Alice.
+See <<Assumptions and goals,assumptions point 3>> for further explanation.
+
+The `swap_digest` field is a https://github.com/multiformats/multihash[multihash-encoded] root hash of a hash list that contains all details of the swap.
+It remains to be defined, how the digest is exactly computed but the idea is the following:
+
+. Take all of the swaps information _known to the COMIT node_ (alpha ledger, alpha asset, protocols, etc) and arrange it in a list
+. Serialize each item into a yet-to-be defined representation
```
bonomat, line 187: Maybe unpopular opinion: why not use protobuf?

If we have _single type messages where its shape is hardcoded by the client_, what do you think of using protobuf and serialize it to binary.

0023-communication-protocol.adoc
bonomat: Something else comes to my mind when reading the proposed messages further down below: 

Do you foresee a `request_information` message as well? 
This could be useful in future cryptographic protocols where an actor can ask for additional information, e.g. "hey, I need a pubkey from you now to do X", "ok, now I need a second different pubkey to do Y", "I need a randomized secret hashed with sha256", ...

It's probably too early to think about this but maybe worth to keep in mind and not exclude it completely from the beginning.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+
+[source,json]
+----
+{
+  "swap_id": "713b0c93-53a1-46b2-9418-de8d46750b79"
+}
+----
+
+In case Bob knows about a swap with the same digest, he generates a UUIDv4 and sends it back to Alice.
+If he doesn't know about a swap with this digest, he closes the substream without sending any message.
```
bonomat, line 221: Let my rephrase to see if I understand the process correctly: 

* Alice adds to her cnd some swap information
* Bob does the same thing
  * both generate the `swap_digest`
* Alice sends to Bob this `swap_digest`
* If Bob knows the `swap_digest` he generates a UUIDv4 (point 3.1.2 Confirmation message) and returns it to Alice.
* ...

Isn't there a danger of a race condition where Alice sends something to Bob but Bob does not expect anything from Alice yet?

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+
+[source,json]
+----
+{
+  "swap_id": "713b0c93-53a1-46b2-9418-de8d46750b79"
+}
+----
+
+Both parties are expected to use this protocol after they have sent all the data they consider necessary for starting the execution of the swap.
+Once each party sent a finalize message AND received one, implementations should consider the swap finalized and start with the execution phase (i.e. watch for transactions, emit actions, etc)
```
bonomat, line 344: I assume that finalize should involve some finale verification? 
Let's say Alice sends `finalize` to Bob, he verifies if he has all information he needs and then returns `finalize` to Alice.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+
+== Discussion points
```
bonomat: _stopped here with the review_

bonomat (commented):
Good job, I like the idea of single type messages. 

@thomaseizinger : as soon as you are happy with this spike. Can you organize a meeting to quickly present the core concepts of this spike. I think everyone of the team should have a clear picture of the implications of the proposed changes before tickets end up in grooming.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+
+This section includes discussion topics that remain to be answered.
+
+=== Should we use multihash for the swap digest
+
+<<Announce-message,Section 3.2>> introduces the concept of a swap digest.
+In order to be forwards compatible with different hash algorithms, it proposes to encode the root hash as a multihash.
+That way, the hash is self-describing and the receiving party knows, which hash algorithm needs to be used to compute the digest.
+
+Do we need the flexibility of changing the hash function that is used to compute the swap digest without a breaking change to the protocol?
```
bonomat, line 356: I understand that the swap digest is independent from the remaining messages and is used for the following things: 
* uniquely announce a swap (provided the function used to compute the digest is collision free)
* allow to verify known information against the swap digest:
  * detect if data is invalid 
  * detect if data is missing

That being said, there might be even better data structures then a multi-hash.

If the swap digest is independent of the other messages we can _play around_ with it and see what works well for us?

0023-communication-protocol.adoc
thomaseizinger: Interactions here refers to interaction with the user (like waiting for the user to send `accept` from the API).

0023-communication-protocol.adoc
thomaseizinger: Actually, I think the original is correct but this sentence uses triple negation. I will fix it, sorry for that.

0023-communication-protocol.adoc
thomaseizinger: I find JSON much easier to work with. If we want binary, we can use CBOR.

PS: I am not the only one disliking protobuf :)
https://reasonablypolymorphic.com/blog/protos-are-wrong/

0023-communication-protocol.adoc
thomaseizinger: Yeah I thought about this (and potentially reversing the protocols). I would follow the design of the `identity` protocol here and maybe create protocols like: `/comit/identity/bitcoin/request/1.0.0` or something like that.

0023-communication-protocol.adoc
thomaseizinger: Your understanding is correct.

> Isn't there a danger of a race condition where Alice sends something to Bob but Bob does not expect anything from Alice yet?

Yes that exists but I am not sure it can be solved completely.
Bob can wait for some time before rejecting the swap but it shouldn't be too long.
Generally, whatever lead the application to instruct cnd to do a swap, the same thing should happen on the other side.

0023-communication-protocol.adoc
```diff
@@ -0,0 +1,383 @@
...
+It is still to be defined, how exactly the swap digest will be computed.
+In particular, we need to define:
+
+* the order in which the tree of information about the swap is arranged
+* which parts of the swap are included in the hash list
+
+Especially the last point is important to discuss.
+Consider the following:
+
+Given the above protocols, it is should be possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
```
da-kami:
```suggestion
Given the above protocols, it is possible to realize an implementation of COMIT that only requires a single user interaction (on both sides) before the swap starts.
```

0023-communication-protocol.adoc
thomaseizinger: No, finalize doesn't involve any information.
The verification was done using the swap_digest.

Regarding the identities, I think it is reasonable to assume that they didn't mess them up.

The reasoning for the finalize message is to make sure, both parties know that there are not going to be any more messages.

da-kami (approved):

0023-communication-protocol.adoc
bonomat: Ah, misunderstood you. Thanks for the clarification.

0023-communication-protocol.adoc
thomaseizinger: > uniquely announce a swap (provided the function used to compute the digest is collision free)

The function is unlikely to be totally collision free because it is oriented around the "offer" that both parties gave to cnd (offer ~ swap - identities).
If we include the cnd identity of the other party within the digest, the only collisions that can happen are between those two parties. If that happens, it is unlikely to be a collision but actually an announcement of a second swap (with the same parameters).

> allow to verify known information against the swap digest:

Yes. The main purpose is to verify information without having to send a lot of data.

> That being said, there might be even better data structures then a multi-hash.

Note that the question you commented on specifically questions the use of the multihash. The alternative would be an untagged hash (where we hardcode the hash function).

> If the swap digest is independent of the other messages we can play around with it and see what works well for us?

Yes that is the idea of splitting up the messages. We should be able to change one part without affecting the others.

thomaseizinger:
> Can you organize a meeting to quickly present the core concepts of this spike. I think everyone of the team should have a clear picture of the implications of the proposed changes before tickets end up in grooming.

@bonomat will do once it is merged!

0023-communication-protocol.adoc
bonomat: Sorry, I wasn't clear: I meant that Bob verifies that he has all information *he needs from Alice* before answering `finalize` to Alice.
 
The protocol proposed works the other way around so far: Alice tells Bob what she thinks he would need. If there is however a problem or a mismatch (an error occurred, protocol update, etc), then it might be the case that Bob would need more information from Alice. In this case, he should nor answer `finalize` to Alice.

0023-communication-protocol.adoc
thomaseizinger: I've revised this section.

0023-communication-protocol.adoc
thomaseizinger: Done!

0023-communication-protocol.adoc
thomaseizinger: I added a section about collision resistance.

0023-communication-protocol.adoc
bonomat: Should Bob in this case indicate that he doesn't know what Alice is talking about? Meaning, return something?

tcharding (approved):

0023-communication-protocol.adoc
thomaseizinger: > Alice tells Bob what she thinks he would need.

Not quite: Both parties tell the other parties what they think they will need.

> Sorry, I wasn't clear: I meant that Bob verifies that he has all information he needs from Alice before answering finalize to Alice.

This is what should happen yes.
The idea is that each party doesn't send `finalize` until:

- they sent all the data they intend to send
- they received all the data they need

If the finalize handshake doesn't happen within a certain time period, we should timeout and abort the swap (there seems to be a bug in the implementations then).

thomaseizinger: Note that the new protocol doesn't have request-reply but just one-way messages. They also all happen on different substreams, meaning it all happens concurrently.

0023-communication-protocol.adoc
thomaseizinger: For example?

A string that Alice can print to the logs?
Should she be able to automatically recover from that?

That topic is a whole new can of worms.
My personal preference would be to evaluate that the need for this as we go and use it ourselves and have other people use it.
That should tell us about common error scenarios and we can think about how to attack them.

I see the general problem, but I don't have a feasible solution at this stage.

thomaseizinger:
I think all of the points are resolved, I will merge this and organize a presentation meeting for next week.

Happy to answer more questions there.

0023-communication-protocol.adoc
thomaseizinger: I just realized that you commented on the digest computation here.

I wouldn't use JSON for that but neither protobuf. The serialization is going to be very simple, like a single key-value which is then hashed.

An example representation could be:

```rust
format!("alpha_ledger={}", alpha_ledger)
```

and then hashing it.

I don't see a point in using a technology like protobuf here, we don't have any structs or anything which kind of defeats the point of an IDL.

Note: Very likely, we don't want to go with the above example because it is string-encoding dependent but rather define our own binary encoding.

PR #52, opened 2020-02-03
Modify section on 'hold invoices'

tcharding:
The HAN-HALight / HALight-HAN swap protocols are not symmetric, make
this clear in the spike.

### DIFF
--- a/0022-lnd-plan-of-attack.adoc
+++ b/0022-lnd-plan-of-attack.adoc
@@ -97,8 +97,11 @@ There is no plan to monitor the underlying blockchain for this.
 
 ==== Standard LN does not support payments without the other party knowing the pre-image
 
-There is an experimental feature called "hold-invoice" which we can utilize to design a symmetric protocol.
-As a consequence, users will have to compile LND with the feature flag "invoicesrpc" to get the desired functionality.
+HAN-HALight swaps can be achieved using standard Lightning.
+HALight-HAN swaps are not possible using standard Lightning because in the COMIT protocol the preimage is known to Alice and not Bob where as for a lightning payment from Alice to Bob the preimage is known to Bob.
+There is an experimental feature called "hold-invoice" which we can utilize to design a protocol that supports HALight-HAN swaps.
+In order to support HALight-HAN swaps both users will have to use lightning nodes executing binaries based on the https://github.com/lightningnetwork/lnd/ project and additionally the binaries must be built with the feature flag "invoicesrpc" to get the desired functionality.
+The steps needed to complete a HALight-HAN swap are _not_ the same as for a HAN-HALight swap.
 
 ==== Our current REST API endpoints are not capable of expressing the concept of a protocol per ledger
 


0022-lnd-plan-of-attack.adoc
```diff
@@ -97,8 +97,11 @@ There is no plan to monitor the underlying blockchain for this.
 
 ==== Standard LN does not support payments without the other party knowing the pre-image
 
-There is an experimental feature called "hold-invoice" which we can utilize to design a symmetric protocol.
-As a consequence, users will have to compile LND with the feature flag "invoicesrpc" to get the desired functionality.
+HAN-HALight swaps can be achieved using standard Lightning.
```
D4nte, line 100: Yes, but actually @thomaseizinger suggested to still use the hodl invoice future for some reasons. The reasons are fair to me, I just don't remember them. It's actually written somewhere in the spike but I suggested @thomaseizinger to state it more explicitely, here would be a good place to do so.

bonomat: As the proposed protocol is symmetric it simplifies the development and results in the same user behaviour in both directions.

We can optimize it at a later stage.

bonomat (changes requested):
@tcharding : we were aware of this limitation and were really happy when LND announced the HoldInvoice feature. Finally we can do both directions.
That was the whole purpose of this spike. :)

0022-lnd-plan-of-attack.adoc
D4nte: > As the proposed protocol is symmetric it simplifies the development and results in the same user behaviour in both directions.

There were more than that I believe.

0022-lnd-plan-of-attack.adoc
```diff
@@ -97,8 +97,11 @@ There is no plan to monitor the underlying blockchain for this.
...
+HALight-HAN swaps are not possible using standard Lightning because in the COMIT protocol the preimage is known to Alice and not Bob where as for a lightning payment from Alice to Bob the preimage is known to Bob.
+There is an experimental feature called "hold-invoice" which we can utilize to design a protocol that supports HALight-HAN swaps.
+In order to support HALight-HAN swaps both users will have to use lightning nodes executing binaries based on the https://github.com/lightningnetwork/lnd/ project and additionally the binaries must be built with the feature flag "invoicesrpc" to get the desired functionality.
+The steps needed to complete a HALight-HAN swap are _not_ the same as for a HAN-HALight swap.
```
thomaseizinger, line 104: That is not true.
The steps are conceptually similar to HAN-HAN except that we start with a `add-hold-invoice`.

0022-lnd-plan-of-attack.adoc
thomaseizinger: > @thomaseizinger to state it more explicitely, here would be a good place to do so.

It does say here: https://github.com/comit-network/spikes/blob/master/0022-lnd-plan-of-attack.adoc#222-what-does-the-protocol-look-like

thomaseizinger (changes requested):
Sorry but IMO, this patch is adding more confusion than clarity to this spike.

- The information about Alice knowing the secret is common knowledge in COMIT.
- The feature flag was already mentioned before.
- The statement about the protocol not being symmetric is false.

0022-lnd-plan-of-attack.adoc
tcharding: Sincere apologies @thomaseizinger, I did read that section (the first time I read the spike, before I really understood what was going on) - I did not remember that section and on subsequent readings did not see it somehow.  I finally get it now, for HAN-HALight the 'hold' period is essentially zero since Alice knows the secret but the protocol remains symmetrical.  Thanks for you patience.

tcharding:
I was wrong.

PR #55, opened 2020-02-17
Add section explaining the software choice

### LINKED ISSUE #1838: Plan of attack for integrating LND into cnd
# Problem

Doing an atomic swap with on-chain transactions is constrained by the underlying blockchain. This has several downsides:
* Transactions are bound to the blocktime. This means, an atomic swap between Bitcoin and Ethereum using on-chain transactions would take on average more than 20 minutes provided both parties are fine with 1 confirmation. 
* Transactions are _expensive_ and the fee needs to be calculated ad-hoc according to network congestions.
* The atomic swap is _traceable_ as both on-chain transactions include the same hash.

# Goal

Come up with a _plan of attack_ for adding support for LND in cnd. 
This should be done in accordance of the a new protocol as mentioned in comit-network/RFCs#134

thomaseizinger:
Resolves https://github.com/comit-network/comit-rs/issues/1838.


bonomat (approved):
👍

tcharding (approved):

D4nte (approved):

PR #60, opened 2020-04-15
Add NAT traversal spike

### LINKED ISSUE #1172: Investigate solutions for connecting two cnds behind NAT
# Problem

Currently, the person in the role of cryptographic Alice needs to send the SWAP REQUEST to the person in the role of the cryptographic Bob. This implies that the person taking the role of cryptographic Bob needs to have a port open.

# Goal 

Ideally, we want to allow users to use COMIT without opening a port.

# Recommendation

I don't remember where this list comes from, so treat it with a grain of salt:

Consider the following:
- Implement circuit relay ourselves (see https://github.com/comit-network/comit-rs/issues/1172#issuecomment-575452698)
- hole punching
- UPnP
- how BitTorrent does it
- available standards
- how Skype does it
- rendezvous nodes

Keep in mind that the solution needs to be compatible with libp2p.

rishflab:
Add spike providing a summary of NAT traversal methods. Provide a
recommended solution and a way to implement suggested solution.

Closes https://github.com/comit-network/comit-rs/issues/1172

### DIFF
--- a/0025-nat-traversal.adoc
+++ b/0025-nat-traversal.adoc
@@ -15,7 +15,7 @@ UPnP: Universal Plug and Play
 
 TURN: Traversal Using Relays around NAT
 
-STUN: Session traversal utilities around NAT
+STUN: Session Traversal Utilities for NAT
 
 == Context
 
--- a/0025-nat-traversal.adoc
+++ b/0025-nat-traversal.adoc
@@ -11,7 +11,7 @@ Tracking issue: https://github.com/comit-network/comit-rs/issues/1172[#1172]
 
 NAT: Network Address Translation/Translator
 
-UPnP: Universal plug and play
+UPnP: Universal Plug and Play
 
 TURN: Traversal Using Relays around NAT
 
--- a/0025-nat-traversal.adoc
+++ b/0025-nat-traversal.adoc
@@ -121,7 +121,7 @@ Other peers will be able to dial through the relay peer, which will forward traf
 
 - - Requires a third peer to serve as a relay for communications between Alice and Bob
 
-- - Requires the third peer to know its inbound ports (either through another STUN/TURN or port forwarding)
+- - Requires the third peer to have an externally facing IP address
 
 - - Requires a third peer to serve as a relay for communications between Alice and Bob
 
--- a/0025-nat-traversal.adoc
+++ b/0025-nat-traversal.adoc
@@ -61,7 +61,7 @@ Bitcoin core removed UPnP support due to security concerns: https://bitcoin.org/
 
 - - Requires the user to have a UPnP capable and enabled NAT (UPnP capable routers are fairly common in home networks)
 
-- - Opens up the user to security vulnerabilities arising from the NAT blindly trusting applications and allowing to open externally facing ports
+- - Opens up the user to security vulnerabilities arising from the NAT blindly trusting applications and allowing incoming traffic to connect to externally facing ports
 
 === Reverse connection
 
--- a/0025-nat-traversal.adoc
+++ b/0025-nat-traversal.adoc
@@ -1,0 +1,161 @@
+= Traversal methods for comit nodes behind a NAT
+Rishab Sharma<rishab.sharma@tenx.tech>;
+:toc:
+:revdate: 2020-03-15
+
+NOTE: Author: {authors} +
+Date: {revdate} +
+Tracking issue: https://github.com/comit-network/comit-rs/issues/1172[#1172]
+
+== Context
+
+Currently a  comit node cannot communicate with another dial another comit node that is behind a NAT(network address translation) device.
+Most routers use NAT to allow multiple devices to use the same public IPv4 address.
+A comit node could be running any of these devices and the dialer needs to know how to traverse the NAT to reach the right device.
+
+== Goals
+
+1. Allow a comit node to traverse a NAT to reach another comit node.
+
+
+== Abbreviations
+
+NAT: Network address translation
+
+UPnP: Universal plug and play
+
+TURN: Traveral using relays around NAT
+
+STUN: Session traversal utilities around NAT
+
+ICE: Interactive connectivity Establishment
+
+== NAT Traversal Methods
+
+This sections provides an overview of current NAT traversal methods. Alice is the dialer and Bob is the receiver behind a NAT device.
+
+=== Manual Port Forwarding
+
+Bob configures his router to forward an externally facing port to his device.
+Bob notifies peers of his public IP and forwarded port so that they can dial him.
+This solution does not require any changes to the COMIT codebase and places the burden on the user to configure their router.
+Users who cannot access the router, for example users on a corporate or mobile network will not be able to use this method.
+
+- - Requires the user to have access to NAT configuration
+
+- - Requires the user to configure NAT settings
+
+- + Alice and Bob can directly communicate through a direct connection
+
+=== Automatic Port Forwarding using UPnP
+
+There is no single perfect solution to solving the NAT traversal problem.There is no single perfect solution to solving the NAT traversal problem.
+UPnP enabled routers automatically forward ports required by an application.
+An application will make a request to the router to open specific ports.
+The user wont have to manually forward the ports used by the comit node.
+Due to security concerns, UPnP is often disabled.
+
+- - Requires the user to behind have a UPnP capable and enabled router
+
+- - Opens up the user to security vulnerabilities arising from allowing applications to open externally facing ports
+
+- + Alice and Bob can directly communicate through a direct connection
+
+=== Reverse connection
+
+Bob dials Alice who uses the inbound connection to send to data to Bob.
+This technique only works if Alice knows her public IP and port and has provided it to Bob.
+Since Bob is unable to dial Alice, Bob is unable to initiate a swap and must wait to be dialled by a peer.
+
+If both Alice and Bob are behind a NAT a third peer which is not behind NAT can be used as a relay.
+Alice and Bob both dial a relay peer. Th
+This technique is used by the Bittorrent network and relies on there being a critical mass of relay peers that have a public IP.
+
+- + Does not require the use to have configure to the NAT settings on their router
+
+- + Does not require the user to have access to NAT configuration
+
+- - Does not require the user to have access to NAT configuration
+
+- - Requires Bob to have a known external IP
+
+
+=== STUN/Hole Punching
+
+Bob sends UDP messages to a third peer(STUN server).
+The STUN server will return which ports and public IP Bobs's UDP messages originated from.
+Bob has discovered public IP and ports which peers can use to dial him.
+Bob
+
+STUN cannot be used to traverse NAT configurations that randomnly assign port and/or IP for new connections.
+A symmetric NAT will randomnly generate a new port and/or public IP for each outbound connection made by Alice.
+This prevents Bob from determining the external IP and ports assigned to her by her NAT.
+Symmetric NAT is commonly used by mobile network providers.
+
+- + Does not require the user to configure NAT settiThere is no single perfect solution to solving the NAT traversal problem.ngs
+
+- + Alice and Bob can directly communicate through a direct connection
+
+- - Requires a third peer to facilitate a connection between two peers
+
+- - Cannot traverse symmetric, cone and restricted cone type NATs
+
+- - Requires a third peer to start a connection between Alice and Bob
+
+- - Requires the third peer to know its inbound ports (either through another STUN/TURN or port forwarding)
+
+=== TURN
+
+If STUN cannot be used to traverse a NAT, a relay peer can be used a relay between the comit_nodes.
+When Bob isn’t able to listen on a public address, Bob can dial out to a relay peer, which will keep a long-lived connection open.
+Other peers will be able to dial through the relay peer, which will forward traffic to Bob.
+
+- + Can be used to traverse all commonly know NAT configurations
+
+- + Does not require the use to have configure NAT settings
+
+- - Requires a third peer to serve as a relay for communications between Alice and Bob
+
+- - Requires the third peer to know its inbound ports (either through another STUN/TURN or port forwarding)
+
+
+== Recommendation
+
+Manual and automatic port forwarding solve the NAT traversal problem through hardware configuration.
+We cannot rely on all users to have access to their network hardware and therefore we will need to implement alternative NAT traversal techniques.
+
+Reverse connections cannot be used because many of our protocols require both peers to be able to initate a connection with each other.
+This means that COMIT protocols will have to redesigned to use the connection initiated by dialer to complete a SWAP.
+Peers behind a NAT device will be unable to play an active role in the COMIT network.
+
+In order to allow comit nodes to traverse any NAT configuration the TURN protocol will need to be implemented.
+ICE recommends that peers attempt to connect using STUN if possible only use TURN a connection using STUN cannot be established.
+STUN is preferred over TURN for performance and efficiency reasons as it does require a third party to relay communications between nodes.
+
+Although TURN servers can resolve the NAT traversal problem, at least one STUN or TURN server that knows its own external ports and IP will be required to bootstrap the network and to restart the network in case it goes down.
+
+Users who have knowledge of their public IP and ports should have the option to run as STUN and/or TURN servers.
+
+
+=== Implementation
+
+libp2p-rs contains an identify protocol which implements functionality similar to the STUN protocol.
+The identify protocol allows the the peer behind a NAT to identify their ports and push that information to other peers.
+
+libp2p contains a circuit relay module based off the TURN protocol.
+There is no Rust implementation of the circuit relay module but the Javascript and Go implementations as an implementation guide.
+
+libp2p also contains a spec for a peer discovery protocol called rendezvous which can be used discover circuit relays or TURN servers.
+It is unclear how the rendezvous will interact with decentralised reputation and orderbook systems as they have no been designed.
+It is not neccesarry or advisable to implement the rendezvous protocol to traverse NAT devices.
+
+== Sources
+
+- Wikipedia article on common NAT classifications: https://en.wikipedia.org/wiki/Network_address_translation#Methods_of_translation
+- Twilio overview of STUN, TURN and ICE protocols for NAT traversal: https://www.twilio.com/docs/stun-turn/faq
+- IPFS tutorial on libp2p's circuit relay: https://github.com/ipfs/js-ipfs/blob/master/examples/circuit-relaying/README.md
+- libp2p circuit relay spec:  https://github.com/libp2p/specs/tree/master/relay
+- libp2p rendezvous spec: https://github.com/libp2p/specs/tree/master/rendezvous
+- libp2p identify spec: https://github.com/libp2p/specs/tree/master/identify
+
+
--- a/0025-nat-traversal.adoc
+++ b/0025-nat-traversal.adoc
@@ -13,7 +13,7 @@ NAT: Network address translation/translator
 
 UPnP: Universal plug and play
 
-TURN: Traveral using relays around NAT
+TURN: Traversal Using Relays around NAT
 
 STUN: Session traversal utilities around NAT
 
@@ -170,5 +170,4 @@ This would reduce the load on TURN servers and centralisation of the network.
 - libp2p circuit relay spec:  https://github.com/libp2p/specs/tree/master/relay
 - libp2p identify spec: https://github.com/libp2p/specs/tree/master/identify
 - Nat traversal discussion thread for go-libp2p: https://github.com/libp2p/go-libp2p/issues/375
-
 
--- a/0025-nat-traversal.adoc
+++ b/0025-nat-traversal.adoc
@@ -9,7 +9,7 @@ Tracking issue: https://github.com/comit-network/comit-rs/issues/1172[#1172]
 
 == Abbreviations
 
-NAT: Network address translation/translator
+NAT: Network Address Translation/Translator
 
 UPnP: Universal plug and play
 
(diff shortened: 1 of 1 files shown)


0025-nat-traversal.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+- - Requires the user to have access to NAT configuration
+
+- - Requires the user to configure NAT settings
+
+- + Alice and Bob can directly communicate through a direct connection
+
+=== Automatic Port Forwarding using UPnP
+
+There is no single perfect solution to solving the NAT traversal problem.There is no single perfect solution to solving the NAT traversal problem.
```
bonomat: Something is odd here :D 
```suggestion
There is no single perfect solution to solving the NAT traversal problem.
```

thomaseizinger:
Fixed your PR description to use a full link instead of just the number, otherwise it is not linked :)

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,161 @@
...
+The STUN server will return which ports and public IP Bobs's UDP messages originated from.
+Bob has discovered public IP and ports which peers can use to dial him.
+Bob
+
+STUN cannot be used to traverse NAT configurations that randomnly assign port and/or IP for new connections.
+A symmetric NAT will randomnly generate a new port and/or public IP for each outbound connection made by Alice.
+This prevents Bob from determining the external IP and ports assigned to her by her NAT.
+Symmetric NAT is commonly used by mobile network providers.
+
+- + Does not require the user to configure NAT settiThere is no single perfect solution to solving the NAT traversal problem.ngs
```
bonomat:
```suggestion
- + Does not require the user to configure NAT settings. There is no single perfect solution to solving the NAT traversal problem.
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,161 @@
...
+This means that COMIT protocols will have to redesigned to use the connection initiated by dialer to complete a SWAP.
+Peers behind a NAT device will be unable to play an active role in the COMIT network.
+
+In order to allow comit nodes to traverse any NAT configuration the TURN protocol will need to be implemented.
+ICE recommends that peers attempt to connect using STUN if possible only use TURN a connection using STUN cannot be established.
+STUN is preferred over TURN for performance and efficiency reasons as it does require a third party to relay communications between nodes.
+
+Although TURN servers can resolve the NAT traversal problem, at least one STUN or TURN server that knows its own external ports and IP will be required to bootstrap the network and to restart the network in case it goes down.
+
+Users who have knowledge of their public IP and ports should have the option to run as STUN and/or TURN servers.
```
bonomat: I'm not sure if I follow :) 

Do you propose that we provide both solutions (STUN and TURN)? 

Can you rephrase this section so that it is clear what your recommendation is please? Same goes for the next section.

bonomat (commented):
Thanks for writing this up. Can you rephrase the recommendation section please? For me it is not clear what we should do to solve this problem.

There are some weird errors in the text, can you fix them please? I higlighted only a few but there are more.

0025-nat-traversal.adoc
rishflab: woops

0025-nat-traversal.adoc
rishflab: yes we should ideally provide both solutions if want to be able to traverse most known NAT configurations. I'll reword this section to make it clear

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,161 @@
...
+- + Can be used to traverse all commonly know NAT configurations
+
+- + Does not require the use to have configure NAT settings
+
+- - Requires a third peer to serve as a relay for communications between Alice and Bob
+
+- - Requires the third peer to know its inbound ports (either through another STUN/TURN or port forwarding)
+
+
+== Recommendation
```
thomaseizinger, line 132: This sounds more like a summary to me :)

thomaseizinger (changes requested):
This spike gives a good overview of the problem but lacks some depth in terms of its recommendation in my opinion.

I've compiled a list of questions that you may find helpful to go into more depth: (@bonomat if you disagree with the level of detail on some of these, please speak up)

- If we make use of the libp2p protocols like identify and circuit-relay, what does the communication pattern look like? Can we have a sequence diagram for this maybe?
- How does the circuit-relay solution actually work? To my knowledge this is a dedicated libp2p protocol. How can we open a substream to a third node via a circuit-relay node?
- Do we need a fixed set of bootnodes for this?
	- Are there ready-to-use circuit-relay instances that we can use?
	- If no, can we set one up using the Go implementation or are there compatibility issues?
- Is the identify + circuit-relay implementation transparent to the user?
Currently, we pass a PeerID and a multi-addr to `cnd` if we want to initiate a swap. Is that enough?

Personally, I would probably to a small PoC to answer all these questions :)

rishflab:
> This spike gives a good overview of the problem but lacks some depth in terms of its recommendation in my opinion.
> 
> I've compiled a list of questions that you may find helpful to go into more depth: (@bonomat if you disagree with the level of detail on some of these, please speak up)
> 
> * If we make use of the libp2p protocols like identify and circuit-relay, what does the communication pattern look like? Can we have a sequence diagram for this maybe?
> * How does the circuit-relay solution actually work? To my knowledge this is a dedicated libp2p protocol. How can we open a substream to a third node via a circuit-relay node?
> * Do we need a fixed set of bootnodes for this?
>   
>   * Are there ready-to-use circuit-relay instances that we can use?
>   * If no, can we set one up using the Go implementation or are there compatibility issues?
> * Is the identify + circuit-relay implementation transparent to the user?
>   Currently, we pass a PeerID and a multi-addr to `cnd` if we want to initiate a swap. Is that enough?
> 
> Personally, I would probably to a small PoC to answer all these questions :)

Just pushed a comit that improved the spike and just saw your comment. Happy to look into this further tommorow but I am not sure if its within the scope of the ticket.

thomaseizinger:
> Just pushed a comit that improved the spike and just saw your comment. Happy to look into this further tommorow but I am not sure if its within the scope of the ticket.

Happy to discuss the scope but I would argue that I am not increasing it but rather clarifying it :)

If I had to implement a solution for NAT traversal for cnd now, I would still have to read up and basically answer some/all of the above questions so I feel the spike is not going deep enough :)

More specifically, I am lacking the focus on libp2p. It is good to know and learn about TURN, STUN, ICE etc but how does this help us in practice? We don't have control over the low-level network connection so any solution that requires that is not possible I think.

I hope my feedback is not discouraging! It is definitely not an easy job to write a good spike as you have to balance how much research you are doing and which questions you are going to answer.

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+- - Opens up the user to security vulnerabilities arising from the NAT blindly trusting applications and allowing to open externally facing ports
+
+=== Reverse connection
+
+Instead of Alice attempting to reach Bob through his NAT. Bob dials Alice and Alice uses the connection to send to data to Bob.
+This technique only works if Alice knows her public IP and port and has provided it to Bo.
+Since Alice is unable to dial Bob, Alice is unable to initiate communication and must wait to be dialled by a Bob.
+
+Reverse connections cannot be used because many of our protocols require both peers to be able to initate a connection with each other.
```
thomaseizinger, line 72: That is not quite true.

Libp2p is specifically designed so you can open substreams in either direction without having to care about the physical connection.

If for some reason we already have a physical connection to a node, then any of the peers can use this with any kind of protocol on top.

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+- - Requires a third peer to serve as a relay for communications between Alice and Bob
+
+- - Requires peers to know peers that are TURN servers
+
+- - Leads to some degree of centralisation
+
+== Recommendation
+
+Assumption: A significant portion of comit nodes will have forwarded ports.
```
thomaseizinger, line 134: I think the opposite is going to be true unfortunately.

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+We should implement the following NAT traversal methods in the order listed below.
+After each implementation we should evaluate whether it is necessary to implement the subsequent NAT traversal methods.
+
+1. TURN (essential for maximum NAT coverage)
+2. STUN (optional to reduce load and reliance on TURN servers)
+3. UPnP (optional to reduce load and reliance on TURN servers)
+
+TURN is the first prioirty as it is capable of traversing most commonly known NAT configurations.
+In order for TURN to work, some port forwarded peers will have to operate as TURN servers to allow NAT restricted peers to communicate.
```
thomaseizinger, line 144: Can we host dedicated TURN servers?
To my understanding, those can operate solely on libp2p and don't have to know anything about COMIT. Is that correct?

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+The libp2p specification defines a circuit relay module based off the TURN protocol.
+This protocol allows NAT restricted nodes to "proxy" through other p2p nodes that volunteer to act as p2p-circuit relays.
+These nodes will announce to the network that they are accepting relay connections using libp2p's peer routing.
```
thomaseizinger, line 148: How do we get started initially? i.e. how does a node behind NAT know about these relay nodes?

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+STUN would be implemented if we found that there are insufficient nodes willing to act as relays due bandwidth concerns.
+STUN allows peers to discover their public IP and ports and establish a communicate directly without a relay.
+
+libp2p has it's own version of the STUN protocol using peer-routing, identify, and reuseport.
+
+1. reuseport uses the same source port for both dialing and listening allowing us to accept new inbound connections on the source port used by the router
+2. identify receives observed address information from all connected peers
+3. peer-routing allows us to publish these addresses to the network and allow peers to discover their own public address on which they can accept inbound connections
+
+If both STUN and TURN are implemented, STUN will be the preferred method and TURN will be used as a fallback if the public address cannot be determined or predicted using STUN.
```
thomaseizinger, line 160: Given that COMIT is very much a 1on1 application at the moment, I think the preferred solution should be something where a node can connect to another one out of the blue. What would you propose for that?

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+Date: {revdate} +
+Tracking issue: https://github.com/comit-network/comit-rs/issues/1172[#1172]
+
+== Abbreviations
+
+NAT: Network address translation/translator
+
+UPnP: Universal plug and play
+
+TURN: Traveral using relays around NAT
```
tcharding:
```suggestion
TURN: Traversal Using Relays around NAT
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+:toc:
+:revdate: 2020-03-15
+
+NOTE: Author: {authors} +
+Date: {revdate} +
+Tracking issue: https://github.com/comit-network/comit-rs/issues/1172[#1172]
+
+== Abbreviations
+
+NAT: Network address translation/translator
```
tcharding:
```suggestion
NAT: Network Address Translation/Translator
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+UPnP: Universal plug and play
```
tcharding:
```suggestion
UPnP: Universal Plug and Play
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+TURN: Traveral using relays around NAT
+
+STUN: Session traversal utilities around NAT
```
tcharding:
```suggestion
STUN: Session Traversal Utilities for NAT
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+Not all NATs are UPnP capable and it is often disabled or not used by peer to peer applications.
+Bitcoin core removed UPnP support due to security concerns: https://bitcoin.org/en/alert/2015-10-12-upnp-vulnerability.
+
+- + Does not require Bob to configure his NAT
+
+- + Alice and Bob can communicate through a direct connection
+
+- - Requires the user to have a UPnP capable and enabled NAT (UPnP capable routers are fairly common in home networks)
+
+- - Opens up the user to security vulnerabilities arising from the NAT blindly trusting applications and allowing to open externally facing ports
```
tcharding:
```suggestion
- - Opens up the user to security vulnerabilities arising from the NAT blindly trusting applications and allowing incoming traffic to connect to externally facing ports
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+=== Reverse connection
+
+Instead of Alice attempting to reach Bob through his NAT. Bob dials Alice and Alice uses the connection to send to data to Bob.
```
tcharding, line 68:
```suggestion
Instead of Alice attempting to reach Bob through his NAT. Bob dials Alice and Alice uses the connection to send data to Bob.
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+This technique only works if Alice knows her public IP and port and has provided it to Bo.
```
tcharding, line 69:
```suggestion
This technique only works if Alice knows her public IP and port and has provided it to Bob.
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+Since Alice is unable to dial Bob, Alice is unable to initiate communication and must wait to be dialled by a Bob.
```
tcharding, line 70:
```suggestion
Since Alice is unable to dial Bob, Alice is unable to initiate communication and must wait to be dialled by Bob.
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+- - Alice can not dial Bob and therefore cannot initiate a swap or use Bob as a peer to propagate information to the network
+
+- - We will most likely need to redesign protocols to utilise reverse connections
+
+- - Requires Bob to have a known external IP
+
+=== STUN/Hole Punching
+
+Bob sends UDP messages to a third peer known as a STUN server.
```
tcharding, line 90:
```suggestion
Bob sends UDP messages to a third party known as a STUN server.
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+The STUN server will return which ports and public IP Bobs's UDP messages originated from.
```
tcharding, line 91:
```suggestion
The STUN server will return which port and IP Bobs's UDP message originated from.
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+Bob has now discovered public IP and ports which peers can use to dial him.
```
tcharding, line 92:
```suggestion
Bob has now discovered a public IP and port which peers can use to dial him.
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+Bob can provide this knowledge to Alice, allowing Alice to dial him.
```
tcharding, line 93: Not entirely true, even with an IP and port, if this is behind NAT the port will need to be forwarded to the local machine's IP address.  Might be worth mentioning this quote from wikipedia
> STUN is not a self-contained NAT traversal solution applicable in all NAT deployment scenarios and does not work correctly with all of them. It is a tool among other methods and it is a tool for other protocols in dealing with NAT traversal, most notably Traversal Using Relay NAT (TURN) and Interactive Connectivity Establishment (ICE).

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+- - Requires peers to know peers that are STUN servers
+
+- - Leads to some degree of centralisation
+
+=== TURN
+
+When Bob isn’t able to determine or predict his public facing IP address and port, Bob can dial out to a relay peer, which will keep a long-lived connection open.
+Other peers will be able to dial through the relay peer, which will forward traffic to Bob.
+
+- + Can be used to traverse all commonly know NAT configurations
```
tcharding, line 118:
```suggestion
- + Can be used to traverse all commonly known NAT configurations
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+- + Does not require the Bob to configure his NAT
```
tcharding:
```suggestion
- + Does not require Bob to configure his NAT
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+- - Requires a third peer to serve as a relay for communications between Alice and Bob
+
+- - Requires the third peer to know its inbound ports (either through another STUN/TURN or port forwarding)
```
tcharding:
```suggestion
- - Requires the third peer to have an externally facing IP address
```

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+== Recommendation
+
+Assumption: A significant portion of comit nodes will have forwarded ports.
+
+We should implement the following NAT traversal methods in the order listed below.
+After each implementation we should evaluate whether it is necessary to implement the subsequent NAT traversal methods.
+
+1. TURN (essential for maximum NAT coverage)
+2. STUN (optional to reduce load and reliance on TURN servers)
+3. UPnP (optional to reduce load and reliance on TURN servers)
```
tcharding, line 141: We don't have to 'implement' anything to get this right, just make public the ports we use.

0025-nat-traversal.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+== Recommendation
+
+Assumption: A significant portion of comit nodes will have forwarded ports.
+
+We should implement the following NAT traversal methods in the order listed below.
+After each implementation we should evaluate whether it is necessary to implement the subsequent NAT traversal methods.
+
+1. TURN (essential for maximum NAT coverage)
+2. STUN (optional to reduce load and reliance on TURN servers)
```
tcharding, line 140: As suggested above, I don't think STUN is a standalone solution.

0025-nat-traversal.adoc
tcharding, line 144: Not fully true, as commented above, the servers need an Internet facing IP address.

tcharding (commented):
Good research mate, I'm not convince about the recommendation section in general though.

0025-nat-traversal.adoc
thomaseizinger: Isn't that what he says with "some port forwarded peers"?

0025-nat-traversal.adoc
rishflab: I am guessing most people will be running a node on their home computer. Most routers have UPnP these days and I don't think people disable it.

0025-nat-traversal.adoc
tcharding: The bare minimum for this to work is a publicly available IP address, how one gets it (i.e., by port forwarding) is immaterial and should not be discussed as it just adds noise (IMNSHO).

0025-nat-traversal.adoc
thomaseizinger: If that is true, that would be good news! Do you by any chance have any sources for these claims? :)
The UPnP part, not about where users are going to run COMIT :D

0025-nat-traversal.adoc
rishflab: Yes

0025-nat-traversal.adoc
rishflab: The simplest solution would be to allow the user to specify a list of relay nodes through a config file or REST API call. We could provide a default list of nodes that we approve of similar to bootstrap nodes in Ethereum.

To further improve on this solution you could ask your "friends" or peers that you trust for a list of relay nodes that they trust. At this stage we are entering the realm of decentralised trust systems. The general problem becomes: "How do I know which peers are trustworthy providers of information?".  The information could be a list of reliable relay nodes or orders from an orderbook. I don't think this an easy problem to address and I don't think a potential solution to this problem should be a part of this spike. I think solving this problem will also solve the or partially solve the decentralised orderbook problem.

0025-nat-traversal.adoc
rishflab: Without a TURN server or some form of third to relay data we cannot guarantee that COMIT nodes running on mobile networks or the more restrictive NAT configurations will be able to communicate. 

There are publicly available STUN servers would allow for most users to their NAT and connect to each other. [public STUN server list](https://gist.github.com/zziuni/3741933). We could provide a default list like Ethereum does with bootstrap nodes or allow the specify their own list via config or REST API.

[Cases when STUN will fail](https://stackoverflow.com/questions/23655243/nat-traversal-probability-of-success-using-stun)

0025-nat-traversal.adoc
rishflab: AFAIK to take advantage of UPnP we have to implement some sort of port mapping module.
The go libp2p implementation has one. Perhaps better sense of it than I can [go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat/blob/master/nat.go)

0025-nat-traversal.adoc
rishflab: STUN is a standalone solution for NAT traversal. It does not provide complete coverage of every NAT config.

0025-nat-traversal.adoc
rishflab: good point tobin

PR #61, opened 2020-04-20
Add label-slow-prs GitHub Action

luckysori:


.github/workflows/label-slow-prs.yml
```diff
@@ -0,0 +1,13 @@
+name: Label slow PRs
+
+on:
+  schedule:
+    - cron: "0 2 * * MON"
```
D4nte, line 5: Why not running it every day?

luckysori: Could be a bit wasteful if we're only looking at the results on a weekly basis...

D4nte: ok

D4nte (approved):

PR #62, opened 2020-04-27
new db schema

da-kami:
Note: To properly render the diagrams the links currently point to specific commits. 

ToDo before merge:
- [x] Squash commits
- [x] Change the links in the `puml` files to point to master
- [x] Change the links in the spike document to point to master

### DIFF
--- a/assets/db_schema/all_entities.iuml
+++ b/assets/db_schema/all_entities.iuml
@@ -1,0 +1,294 @@
+' Zero or One	|o--
+' Exactly One	||--
+' Zero or Many	}o--
+' One or Many	}|--
+' One to Many	||--o{
+' Many to One	}o--||
+
+@startuml(id=swap)
+
+entity "swap" as swap {
+    id : <<pk>>,
+    ---
+    local_swap_id : <<unique>>,
+    role,
+    counter_party_peer_id
+}
+
+entity "shared_swap_id" as shared_swap_id {
+    id : <<pk>> <<fk>>,
+    ---
+    shared_swap_id
+}
+
+entity "address_hint" as address_hint {
+    id : <<pk>>
+    ---
+    counter_party_peer_id : <<unique>>,
+    address_hint
+    }
+
+entity "secret_hash" as secret_hash {
+    swap_id : <<pk>> <<fk>>,
+    ---
+    secret_hash
+}
+
+@enduml
+
+@startuml(id=swap_connections)
+
+shared_swap_id |o--|| swap
+address_hint |o--|{ swap
+secret_hash |o--|| swap
+
+@enduml
+
+@startuml(id=han_bitcoin_swap_protocols)
+
+entity "han_bitcoin_swap_detail" as han_bitcoin {
+    han_bitcoin_swap_detail_id : <<pk>>,
+    ---
+    swap_id_fk : <<fk>>,
+    amount,
+    network,
+    expiry,
+    hash_function,
+    redeem_identity,
+    refund_identity,
+    ledger : (alpha | beta)
+}
+
+@enduml
+
+@startuml(id=han_ethereum_swap_protocols)
+
+entity "han_ethereum_swap_detail" as han_ethereum {
+    han_ethereum_swap_detail_id : <<pk>>,
+    ---
+    swap_id_fk : <<fk>>,
+    amount,
+    network,
+    expiry,
+    hash_function,
+    redeem_identity,
+    refund_identity,
+    ledger : (alpha | beta)
+}
+
+@enduml
+
+@startuml(id=han_bitcoin_swap_protocols_state)
+entity "han_bitcoin_state" as han_bitcoin_states {
+    han_bitcoin_state_id : <<pk>>,
+    ---
+    name:
+    (
+        funded
+        | redeemed
+        | refunded
+        | incorrectly_funded
+    )
+}
+
+entity "han_bitcoin_ledger_state" as han_bitcoin_ledger_states {
+    han_bitcoin_ledger_state_id : <<pk>>,
+    ---
+    han_bitcoin_swap_detail_id : <<fk>>,
+    han_bitcoin_state_id : <<fk>>,
+    tx,
+    timestamp
+}
+
+
+@enduml
+
+
+@startuml(id=han_ethereum_swap_protocols_state)
+entity "han_ethereum_state" as han_ethereum_states {
+    han_ethereum_state_id : <<pk>>,
+    ---
+    name:
+    (
+        funded
+        | redeemed
+        | refunded
+        | incorrectly_funded
+    )
+}
+
+entity "han_ethereum_ledger_state" as han_ethereum_ledger_states {
+    han_ethereum_ledger_state_id : <<pk>>,
+    ---
+    han_ethereum_swap_detail_id : <<fk>>,
+    han_ethereum_state_id : <<fk>>,
+    tx,
+    timestamp
+}
+
+@enduml
+
+@startuml(id=han_bitcoin_swap_protocols_connections)
+
+han_bitcoin |o--|| swap
+
+@enduml
+
+@startuml(id=han_bitcoin_swap_protocols_connections_state)
+
+han_bitcoin_ledger_states }--|| han_bitcoin_states
+han_bitcoin_ledger_states }---|| han_bitcoin
+
+@enduml
+
+@startuml(id=han_ethereum_swap_protocols_connections)
+
+han_ethereum |o--|| swap
+
+@enduml
+
+@startuml(id=han_ethereum_swap_protocols_connections_state)
+
+han_ethereum_ledger_states }--|| han_ethereum_states
+han_ethereum_ledger_states }--|| han_ethereum
+
+@enduml
+
+@startuml(id=herc20_swap_protocols)
+
+entity "herc20_swap_detail" as herc20 {
+    herc20_swap_detail_id : <<pk>>,
+    ---
+    swap_id_fk : <<fk>>,
+    amount,
+    network,
+    expiry,
+    hash_function,
+    token_contract,
+    redeem_identity,
+    refund_identity,
+    ledger : (alpha | beta)
+}
+
+@enduml
+
+@startuml(id=herc20_swap_protocols_state)
+entity "herc20_state" as herc20_states {
+    herc20_state_id : <<pk>>,
+    ---
+    name:
+    (
+        deployed
+        | funded
+        | redeemed
+        | refunded
+        | incorrectly_funded
+    )
+}
+
+entity "herc20_ledger_state" as herc20_ledger_states {
+    herc20_ledger_state_id : <<pk>>,
+    ---
+    herc20_swap_detail_id : <<fk>>,
+    herc20_state_id : <<fk>>,
+    tx,
+    timestamp
+}
+@enduml
+
+@startuml(id=herc20_swap_protocols_connections)
+
+herc20 |o--|| swap
+
+@enduml
+
+@startuml(id=herc20_swap_protocols_connections_state)
+herc20_ledger_states }--|| herc20_states
+herc20_ledger_states }--|| herc20
+@enduml
+
+@startuml(id=halight_swap_protocols)
+
+entity "halight_swap_detail" as halight {
+    halight_swap_detail_id : <<pk>>,
+    ---
+    swap_id_fk : <<fk>>,
+    amount,
+    network,
+    chain,
+    expiry,
+    cltv_expiry,
+    hash_function,
+    redeem_identity,
+    refund_identity,
+    ledger : (alpha | beta)
+}
+
+@enduml
+
+@startuml(id=halight_swap_protocols_state)
+entity "halight_state" as halight_states {
+    halight_state_id : <<pk>>,
+    ---
+    name:
+    (
+        started
+        | opened
+        | accepted
+        | settled
+        | canceled
+    )
+}
+
+entity "halight_ledger_state" as halight_ledger_states {
+    halight_ledger_state_id : <<pk>>,
+    ---
+    halight_swap_detail_id : <<fk>>,
+    halight_state_id : <<fk>>,
+    timestamp
+}
+@enduml
+
+@startuml(id=halight_swap_protocols_connections)
+
+halight |o--|| swap
+
+@enduml
+
+@startuml(id=halight_swap_protocols_connections_state)
+
+halight_ledger_states }--|| halight_states
+halight_ledger_states }--|| halight
+
+@enduml
+
+@startuml(id=swap_states)
+
+entity "swap_communication_state" as swap_communication_state {
+    swap_communication_state_id : <<pk>>,
+    ---
+    name:
+    (
+        announced
+        | finalized
+    )
+}
+
+entity "swap_communication_state_for_swap" as swap_communication_states_joint_table {
+    swap_communication_state_for_swap_id: <<pk>>,
+    ---
+    swap_id : <<fk>>,
+    swap_communication_state_id : <<fk>>
+    timestamp
+}
+
+
+
+@enduml
+
+@startuml(id=swap_states_connections)
+
+swap_communication_states_joint_table }o--|| swap_communication_state
+swap_communication_states_joint_table }o--|| swap
+
+@enduml
--- a/0025-an-extensible-db-design.adoc
+++ b/0025-an-extensible-db-design.adoc
@@ -1,0 +1,285 @@
+= Database design for cnd
+Philipp Hoenisch <philipp@coblox.tech>; Daniel Karzel <daniel.karzel@coblox.tech>;
+:toc:
+:revdate: 2020-04-23
+
+NOTE: Author: {authors} +
+Date: {revdate} +
+Status: Draft
+
+== Problem
+
+Our current database design is complex and hard to maintain or to extend.
+The purpose of this spike is to propose a properly normalized relational database model.
+
+=== Scope
+
+The cnd database is used for the (sole) purpose of allowing recovery and efficiently restarting cnd nodes.
+We do not follow a model-driven approach where the database dictates e.g. what swap combinations are actually possible.
+It is out of scope for this spike to come up with a domain model, the spike should focus only on a model for the relevant data.
+This model, however, should adhere to database normalization constraints.
+
+The following protocols are in scope:
+
+* HAN
+* HErc20
+* HALight
+
+Protocols based on scriptless scripts are not in focus, but known constraints (e.g. no secret) are taken into consideration.
+
+=== Database normalization
+
+Database normalization is a process to organize a database into tables and columns.
+The main idea with this is that a table should be about a specific topic and only supporting topics included.
+If a table is not properly normalized and has data redundancy then it will not only eat up extra memory space but will also make it difficult to handle and update the database, without facing data corruption.
+Insert, Update and Delete anomalies are very frequent if database is not normalized.
+There are three main reasons to normalize a database.
+
+* to minimize duplicate data,
+* to minimize or avoid data modification issues,
+* to simplify queries.
+
+Most database models are free of anomalies when reaching 3 NF footnote:[https://en.wikipedia.org/wiki/Database_normalization]; note that some models may require BCNF (extension of 3NF).
+Normalizing the database model beyond 3 NF is not in scope of this spike but might be considered if deemed necessary.
+
+==== 1 NF:
+* each cell to be a single value
+* entries in a column are of the same type
+* rows are uniquely identified - i.e. each row has a unique ID
+
+==== 2 NF:
+* all attributes (non-key columns) depend on the key - if not, then a conjunction table should be created consisting out of 2 keys joining two otherwise not dependent tables (e.g. protocol and swap)
+
+==== 3 NF:
+* all fields (all columns) can be determined only by the key in the table and no other column. E.g. if for every Han_BTC swap we have the same expiries, then these should be in their own table. --> update anomaly
+
+==== BCNF
+* any part of a primary key must not have a functional dependency on another column of the table.
+
+==== 4 NF:
+* no multi-valued dependencies. E.g. a customer who bought 2 items should not be in a table twice with the bought item in one column
+
+
+== Approach
+
+. Data collection: Collect all the data to be stored.
+. First model: Create a first model to act upon.
+. Iterations: Iterate on the model to improve and normalize it.
+. Final model
+
+Constraints we considered besides the normalization constraints:
+
+. FK  relationships should not be defined in a parent table (problem of not being able to create the parent if the child does not exist)
+. Certain data of a swap may only be known to a specific role at a later point in time (after creation of the swap). We want to avoid table updates.
+
+
+== Data collection
+
+We identified the following data sources needed:
+
+. Initial data of each swap when being created for a specific role for a specific alpha and beta protocol where alpha and beta is one of: `HAN_Bitcoin`, `HAN_Ethereum`, `HErc20`, `HALight`
+. Additional data added to the swap for a specific role (e.g. secret hash only known to Bob during announcement)
+. Additional data added to the swap's alpha or beta ledger data.
+
+While not necessarily in scope we considered the following as well:
+
+. Communication State (one per swap - record all events with timestamp)
+. Ledger State (two per swap, record all events with timestamp and additional information if applicable, e.g. transaction ID)
+
+
+== Final model
+
+Instead of starting from scratch, we started iterating over a slightly adapted table layout based on the one proposed by @bonomat
+https://3.basecamp.com/4403044/buckets/16118249/messages/2583518881#__recording_2600021433[in basecamp].
+
+The final full model can be found at the end of this document.
+Our reasoning for creating certain table relations is summed up in the sections below.
+
+=== Swaps
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/swap.puml&fmt=svg[Swaps]
+
+Our first challenge was depicting `alpha` and `beta` protocols for the swap.
+We found at least 3 different approaches how to model this:
+
+1)
+
+[source]
+----
+- `swap`
+	- id
+    - ...
+
+- `han_ethereum_protocol`
+	- id
+	- wei_quantity
+	- ...
+
+- `han_bitcoin_protocol`
+	- id
+	- satoshi_quantity
+	- ...
+
+- `han_ethereum_han_bitcoin_swap`
+	- id
+	- alpha_ledger_fk -> `han_ethereum_protocol`
+	- beta_ledger_fk -> `han_bitcoin_protocol`
+	- swap_fk -> `swap`
+	- ...
+----
+
+This approach would lead to several tables, i.e. one join table for each ledger combination
+
+2)
+[source]
+----
+- `swap`
+	- id
+    - ...
+
+- `han_ethereum_swap_details`
+	- id
+	- wei_quantity
+	- ...
+
+
+- `alpha_ledger_ethereum`
+	- id
+	- local_swap_id_fk -> `swap`
+	- han_ethereum_swap_details_fk -> `han_ethereum_swap_details`
+	- ...
+
+- `beta_ledger_ethereum`
+	- id
+	- local_swap_id_fk -> `swap`
+	- han_ethereum_swap_details_fk -> `han_ethereum_swap_details`
+	- ...
+----
+
+This approach would lead to  2 additional table per protocol.
+
+3)
+[source]
+----
+
+- `protocol`
+    - id
+    - name -> (han_bitcoin | han_ethereum | ...)
+
+- `swap`
+	- id
+    - alpha_ledger_fk -> protocol_id
+    - beta_ledger_fk -> protocol_id
+    - ...
+
+- `han_ethereum_swap_details`
+	- id
+	- wei_quantity
+	- ...
+
+----
+
+This approach does not give us full type safety as data stored in `han_ethereum_swap_details` is not bound to the swap at all.
+
+4)
+
+[source]
+----
+
+
+- `swap`
+	- id
+    - ...
+
+- `han_ethereum_swap_details`
+	- id
+    - swap_id_fk -> swap_id
+	- wei_quantity
+    - ledger : (alpha | beta)
+	- ...
+
+----
+
+Eventually we decided to go for the last approach as this is the least amount of added tables reduces the join complexity when loading data.
+
+This allows us to :
+
+* add additional protocols easily (one new table per protocol, e.g. `{protocol}_swap_detail`)
+* rename protocol tables as needed :)
+
+The second thing which was a bit challenging was how to model data which is only known at a later point in time and how to model data which not all `Swaps` have in common.
+We decided to go for an _extension table_ approach which allows us to add additional information where needed:
+
+* Bob only learns the hash of the secret during the announcement, additionally not every Swap has a secret hash.
+* The shared swap id is only available after a swap was finalized.
+
+We added `secret_hash` and `shared_swap_id` as extension tables, that have the `swap_local_id` as primary key (so there can only be one entry per swap).
+This allows us to add information to one specific swap without having to update the `swap` table.
+Note that the actual secret is not stored in the database, Alice always computes it from the swap parameters and her seed.
+
+Third we dealt with the problem that there is not always an `address_hint` (i.e. Bob does not need to know the `address_hint` of Alice).
+We added the `address_hint` extension table that relates to `swap` on the `counter_party_peer_id` field (rather than on the `swap_local_id`).
+
+Note that the role could be factored out of the table, but we decided against it because it adds unnecessary complexity for now.
+
+What remains is a simple table for `swap` which contains only the information which all swaps have in common: `local_swap_id` `role` and `counter_party_peer_id`.
+
+=== Protocols
+
+[[HanProtocols]]
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/han_bitcoin_swap_protocols_without_state.puml&fmt=svg[HanBitcoinProtocols]
+
+In order to store details for each swap such as data we need for creating the HTLCs for each protocol we introduce `\{protocol\}_swap_details`, where `\{protocol\}` can be `han_bitcoin`, ``han_ethereum`, `herc20` or `halight` for now.
+Similar to the `Swaps` table, we add information which is only known at a later point of time into a sub-table.
+This resulted in quite an excess of sub-tables such as `han_bitcoin/ethereum_redeem_identity`, `han_bitcoin/ethereum_refund_identity`, etc.
+
+Some notable sub-tables:
+
+* `han_bitcoin/ethereum_swap_detail` contains data needed to create the HTLCs. We decided to give this table a generated primary key, rather than using `swap_local_id` as the primary key to allow same chain swaps.
+* `han_bitcoin_ethereum_redeem/refund_identity`: Since some identities are exchanged during announcement, the identities are kept in extension tables.
+
+HErc20 and HaLight follow the same principles as described above.
+For completeness we depict `herc20` and `halight` diagrams below.
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/han_ethereum_swap_protocols_without_state.puml&fmt=svg[HanEthereumProtocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/herc20_swap_protocols_without_state.puml&fmt=svg[Herc20Protocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/halight_swap_protocols_without_state.puml&fmt=svg[HalightProtocols]
+
+
+==== Protocols with states:
+
+While states are generally out of scope we keep these here for future inspiration:
+
+`han_bitcoin/ethereum_ledger_state`: is a new table which we introduced in this database schema.
+It records observed ledger events such as `fund`/`redeem`/`refund`-transaction.
+We extracted the states into `han_bitcoin_state` in order to reduce duplication in each table and not have an extra column for the event-type (`fund`/`redeem`/`refund`).
+We could reuse this table across all `han_*` protocols but decided for separate _state_ tables per protocol.
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/han_bitcoin_swap_protocols_with_state.puml&fmt=svg[HanBitcoinProtocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/han_ethereum_swap_protocols_with_state.puml&fmt=svg[HanEthereumProtocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/herc20_swap_protocols_with_state.puml&fmt=svg[Herc20Protocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/halight_swap_protocols_with_state.puml&fmt=svg[HalightProtocols]
+
+=== Swap State
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/swap_states.puml&fmt=svg[SwapStates]
+
+Additional to the specific `alpha` and `beta` ledger state (depicted in <<Protocols>>) a `swap` has two additional states:
+
+. the _communication_ state which depicts the state during the announcement phase, i.e. `announced` and `finalized`.
+. an aggregated state of the swap, aggregation of the communication state, ledger states, plus an additional error state, i.e. `in_progress`, `swapped`, `not_swapped`, `internal_failure`
+
+While the former one should be stored in the database, the latter one can be dervived easily from the communication state and the ledger states, hence no additional table is needed.
+
+Note that all `*_state` tables are static tables similar to the `protocol` table.
+These tables contain fixed values that are only changed if the state-model changes.
+
+== Full DB Model
+
+[[COMITDBSchema]]
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/master/assets/0025-database-diagram.puml&fmt=svg[COMITDBSchema]
--- a/README.md
+++ b/README.md
@@ -29,5 +29,6 @@ List of spike outcomes for COMIT network.
 - [spike-0022](0022-lnd-plan-of-attack.adoc) - Plan of attack for integrating LND into cnd
 - [spike-0023](0023-communication-protocol.adoc) - Design a new communication protocol for COMIT
 - [spike-0024](0024-more-protocol-naming.adoc) - Contribution by Natoshi Sakamoto (rejected)
+- [spike-0025](0025-an-extensible-db-design.adoc) - An extensible database design for cnd
 
 For new spike outcomes, please use [template.adoc](template.adoc) as basis.
--- a/assets/0025-database-diagram.puml
+++ b/assets/0025-database-diagram.puml
@@ -1,0 +1,24 @@
+@startuml
+' hide the spot
+hide circle
+
+' avoid problems with angled crows feet
+skinparam linetype ortho
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/swap.puml
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/swap_states.puml
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/han_bitcoin_swap_protocols_with_state.puml
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/han_ethereum_swap_protocols_with_state.puml
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/herc20_swap_protocols_with_state.puml
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/halight_swap_protocols_with_state.puml
+'
+' for local use
+'!include db_schema/swap.puml
+'!include db_schema/swap_states.puml
+'!include db_schema/han_bitcoin_swap_protocols_with_state.puml
+'!include db_schema/han_ethereum_swap_protocols_with_state.puml
+'!include db_schema/herc20_swap_protocols_with_state.puml
+'!include db_schema/halight_swap_protocols_with_state.puml
+
+
+@enduml
--- a/assets/db_schema/halight_swap_protocols_with_state.puml
+++ b/assets/db_schema/halight_swap_protocols_with_state.puml
@@ -1,0 +1,16 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!halight_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!halight_swap_protocols_connections
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!halight_swap_protocols_state
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!halight_swap_protocols_connections_state
+
+' for local use
+'!include all_entities.iuml!halight_swap_protocols
+'!include all_entities.iuml!halight_swap_protocols_state
+'!include all_entities.iuml!halight_swap_protocols_connections
+'!include all_entities.iuml!halight_swap_protocols_connections_state
+
+@enduml
--- a/assets/db_schema/halight_swap_protocols_without_state.puml
+++ b/assets/db_schema/halight_swap_protocols_without_state.puml
@@ -1,0 +1,12 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!halight_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!halight_swap_protocols_connections
+
+' for local use
+'!include all_entities.iuml!halight_swap_protocols
+'!include all_entities.iuml!halight_swap_protocols_connections
+
+@enduml
--- a/assets/db_schema/han_bitcoin_swap_protocols_with_state.puml
+++ b/assets/db_schema/han_bitcoin_swap_protocols_with_state.puml
@@ -1,0 +1,16 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_bitcoin_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_bitcoin_swap_protocols_connections
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_bitcoin_swap_protocols_state
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_bitcoin_swap_protocols_connections_state
+
+' for local use
+'!include all_entities.iuml!han_bitcoin_swap_protocols
+'!include all_entities.iuml!han_bitcoin_swap_protocols_state
+'!include all_entities.iuml!han_bitcoin_swap_protocols_connections
+'!include all_entities.iuml!han_bitcoin_swap_protocols_connections_state
+
+@enduml
--- a/assets/db_schema/han_bitcoin_swap_protocols_without_state.puml
+++ b/assets/db_schema/han_bitcoin_swap_protocols_without_state.puml
@@ -1,0 +1,12 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_bitcoin_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_bitcoin_swap_protocols_connections
+'
+' for local use
+'!include all_entities.iuml!han_bitcoin_swap_protocols
+'!include all_entities.iuml!han_bitcoin_swap_protocols_connections
+
+@enduml
--- a/assets/db_schema/han_ethereum_swap_protocols_with_state.puml
+++ b/assets/db_schema/han_ethereum_swap_protocols_with_state.puml
@@ -1,0 +1,16 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_ethereum_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_ethereum_swap_protocols_connections
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_ethereum_swap_protocols_state
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_ethereum_swap_protocols_connections_state
+
+' for local use
+'!include all_entities.iuml!han_ethereum_swap_protocols
+'!include all_entities.iuml!han_ethereum_swap_protocols_state
+'!include all_entities.iuml!han_ethereum_swap_protocols_connections
+'!include all_entities.iuml!han_ethereum_swap_protocols_connections_state
+
+@enduml
--- a/assets/db_schema/han_ethereum_swap_protocols_without_state.puml
+++ b/assets/db_schema/han_ethereum_swap_protocols_without_state.puml
@@ -1,0 +1,12 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_ethereum_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!han_ethereum_swap_protocols_connections
+'
+' for local use
+'!include all_entities.iuml!han_ethereum_swap_protocols
+'!include all_entities.iuml!han_ethereum_swap_protocols_connections
+
+@enduml
--- a/assets/db_schema/herc20_swap_protocols_with_state.puml
+++ b/assets/db_schema/herc20_swap_protocols_with_state.puml
@@ -1,0 +1,16 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!herc20_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!herc20_swap_protocols_connections
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!herc20_swap_protocols_state
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!herc20_swap_protocols_connections_state
+
+' for local use
+'!include all_entities.iuml!herc20_swap_protocols
+'!include all_entities.iuml!herc20_swap_protocols_state
+'!include all_entities.iuml!herc20_swap_protocols_connections
+'!include all_entities.iuml!herc20_swap_protocols_connections_state
+
+@enduml
--- a/assets/db_schema/herc20_swap_protocols_without_state.puml
+++ b/assets/db_schema/herc20_swap_protocols_without_state.puml
@@ -1,0 +1,12 @@
+@startuml
+' hide the spot
+hide circle
+
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!herc20_swap_protocols
+!include https://raw.githubusercontent.com/comit-network/spikes/master/assets/db_schema/all_entities.iuml!herc20_swap_protocols_connections
+'
+' for local use
+'!include all_entities.iuml!herc20_swap_protocols
+'!include all_entities.iuml!herc20_swap_protocols_connections
+
+@enduml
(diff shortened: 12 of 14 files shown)


0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+==== 1 NF:
+* each cell to be a single value
+* entries in a column are of the same type
+* rows are uniquely identified - i.e. each row has a unique ID
+
+==== 2 NF:
+* all attributes (non-key columns) depend on the key - if not, then a conjunction table should be created consisting out of 2 keys joining two otherwise not dependent tables (e.g. protocol and swap)
+
+==== 3 NF:
+* all fields (all columns) can be determined only by the key in the table and no other column. E.g. if for every Han_BTC swap we have the same expiries, then this should be an own table. --> update anomaly
```
tcharding:
```suggestion
* all fields (all columns) can be determined only by the key in the table and no other column. E.g. if for every Han_BTC swap we have the same expiries, then these should be in their own table. --> update anomaly
```

tcharding:
```suggestion
* all fields (all columns) can be determined only by the key in the table and no other column. E.g. if for every Han_BTC swap we have the same expiries, then these should be in their own table. --> update anomaly
```

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+
+. FK  relationships should not be defined in a parent table (problem of not being able to create the parent if the child does not exist)
+. Certain data of a swap may only be known to a specific role at a later point in time (after creation of the swap). We want to avoid table updates.
+
+
+== Data collection
+
+We identified the following data sources needed:
+
+. Initial data of each swap when being created for a specific role for a specific alpha and beta to protocol where alpha and beta is one of: `HAN`, `HErc20`, `HALight`
```
tcharding:
```suggestion
. Initial data of each swap when being created for a specific role for a specific alpha and beta protocol where alpha and beta is one of: `HAN`, `HErc20`, `HALight`
```

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+Our reasoning for creating certain table relations is summed up in the sections below.
+
+=== Swaps
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/swap.puml&fmt=svg[Swaps]
+
+Our first challange was depicting `alpha` and `beta` protocols for the swap.
+Our first attempt was to not indicate this in the `swap` table but instead just use foreign key relationships.
+For that to work we would need to indicate in the child table whether it is `alpha` or `beta` (see below in `swap_protocol`).
+In order to reduce the amount of duplicated data we decided to add a separate table `protocol` and a `1 to many` relation ship from `swap` to `protocol`.
```
tcharding:
```suggestion
In order to reduce the amount of duplicated data we decided to add a separate table `protocol` and a `1 to many` relationship from `swap` to `protocol`.
```

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+Note that the role could be factored out of the table, but we decided against it because it adds unnecessary complexity for now.
+
+What remains is a simple table for `swap` which contains only the information which all swaps have in common: `role`, `alpha_protocol`, `beta_protocol` and `counter_party_peer_id`.
+
+=== Protocols
+
+[[HanProtocols]]
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/han_swap_protocols.puml&fmt=svg[HanProtocols]
+
+In order to store details for each swap such as data we need for creating the HTLCs for each protocol we introduce `\{protocol\}_swap_details`, where `\{protocol\}` can be `han_bitcoin`, ``han_ethereum`, `hrc20` or `halight` for now.
```
tcharding:
```suggestion
In order to store details for each swap such as data we need for creating the HTLCs for each protocol we introduce `\{protocol\}_swap_details`, where `\{protocol\}` can be `han_bitcoin`, ``han_ethereum`, `herc20` or `halight` for now.
```

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/halight_swap_protocols.puml&fmt=svg[HalightProtocols]
+
+=== Swap State
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/swap_states.puml&fmt=svg[SwapStates]
+
+Additional to the specific `alpha` and `beta` ledger state (depicted in <<Protocols>>) a `swap` has two additional states:
+
+. the _communication_ state which depicts the state during the announcement phase, i.e. `sent`, `accepted`, `rejected`
+. the aggregated state of the swap, aggreaged out of the communication state, ledger states plus on additional error state, i.e. `in_progress`, `swapped`, `not_swapped`, `internal_failure`
```
tcharding:
```suggestion
. the aggregated state of the swap, aggregation of the communication state, ledger states, plus an additional error state, i.e. `in_progress`, `swapped`, `not_swapped`, `internal_failure`
```

tcharding (changes requested):
Nice work, very thorough.  A few minor nits

- Can we use the same terminology as we use in the code base of sharde/local swap ids.  In code we call the `shared_swap_id` and `local_swap_id` but in the proposed database tables we have `swap_shared_id` and `swap_local_id`. (Actually in the final image these terms are mixed up, both are used)

- We have had this discussion before but if we use swap ids from the code base (UUID) for database primary/foreign keys we will have to use strings - this seems to be a bad idea (integer keys being better) but there is some debate online over this. Same goes for using `counter_party_peer_id`.
- Perhaps we should us `herc20_swap_params` instead of `her20_swap_details` to be uniform with terminology in the code?

da-kami:
> Nice work, very thorough. A few minor nits
> 
> * Can we use the same terminology as we use in the code base of sharde/local swap ids.  In code we call the `shared_swap_id` and `local_swap_id` but in the proposed database tables we have `swap_shared_id` and `swap_local_id`. (Actually in the final image these terms are mixed up, both are used)

Note that this was so all table and key names follow the same naming schema. But I don't have a strong opinion here. Generally, I feel, that when implementing this name changes are still OK. I am not sure we have to change the schema because of that, but can.

> * We have had this discussion before but if we use swap ids from the code base (UUID) for database primary/foreign keys we will have to use strings - this seems to be a bad idea (integer keys being better) but there is some debate online over this. Same goes for using `counter_party_peer_id`.

I don't really see a problem here, but we can always add a generated ID. I would see this as implementation detail though, as it does not really concern the schema as such. I would not put that into the spike.

> * Perhaps we should us `herc20_swap_params` instead of `her20_swap_details` to be uniform with terminology in the code?

OK to rename all `detail` to `param`, note that all DB tables are currently singular, so I'd stick to `param` which is a bit weird I feel (because it is not just one parameter in the table).

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+
+== Problem
+
+Our current database design is complex and hard to maintain or to extend.
+The purpose of this spike is to propose a properly normalized relational database model.
+
+=== Scope
+
+The cnd database is used for the (sole) purpose of allowing recovery and efficiently restarting cnd nodes.
+We do not follow a model-driven approach where the database dictates e.g. what swap combinations are actually possible.
```
thomaseizinger: That is interesting, can you expand a bit on why you've decided this way?

Intuitively, I would have said the DB should enforce this but I may be convinced otherwise :)

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+The final model can be found at the end of this document.
+Our reasoning for creating certain table relations is summed up in the sections below.
+
+=== Swaps
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/swap.puml&fmt=svg[Swaps]
+
+Our first challange was depicting `alpha` and `beta` protocols for the swap.
+Our first attempt was to not indicate this in the `swap` table but instead just use foreign key relationships.
+For that to work we would need to indicate in the child table whether it is `alpha` or `beta` (see below in `swap_protocol`).
```
thomaseizinger: Have you considered a model with one join-table per valid swap combination?
Such tables could have a column for alpha and beta ledger that point to a specific protocol table and are hence "type-safe", i.e.:

`han_ethereum_han_bitcoin_swap`:

- id
- fk_alpha_ledger -> han_ethereum
- fk_beta_ledger -> han_bitcoin

What I find interesting about this idea is that the join table could be a good place to add information that belongs to a particular protocol combination, like the secret hash for example.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+In order to reduce the amount of duplicated data we decided to add a separate table `protocol` and a `1 to many` relation ship from `swap` to `protocol`.
+`protocol` is a _static_ table.
+Static means, it contains all available protocols as identifiers.
```
thomaseizinger: Would we ever load this information for something? We've been focusing on expressing these protocols on a type level so there is usually no runtime information in cnd where the protocol is held in a variable.

I get that you need/want this to not duplicate the name in the columns of the `swap` table but I find it kind of odd to have data in a table that we will actually never retrieve, or would we retrieve it for something?

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+* simple update the name if we decide to, i.e. instead of updating X rows in `swaps` we update 1 entry in `protocol`
+* add additional protocols
+
+The second thing which was a bit challenging was how to model data which is only known at a later point in time and how to model data which not all `Swaps` have in common.
+We decided to go for an _extension table_ approach which allows us to add additional information where needed:
+
+* Bob only learns the hash of the secret during the announcement, additionally not every Swap has a secret hash.
+* The shared swap id is only available after a swap was finalized.
+
+We added `secret_hash` and `shared_swap_id` as extension tables, that have the `swap_local_id` as primary key (so there can only be one entry per swap).
```
thomaseizinger: I would recommend to use a `UNIQUE` constraint instead and let the database use IDs for whatever it needs.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+This allows us to add information to one specific swap without having to update the `swap` table.
+Note that the actual secret is not stored in the database, Alice always computes it from the swap parameters and her seed.
```
thomaseizinger: Is the secret hash ever stored for Alice?

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+
+Third we dealt with the problem that there is not always an `address_hint` (i.e. Bob does not need to know the `address_hint` of Alice).
+We added the `address_hint` extension table that relates to `swap` on the `counter_party_peer_id` field (rather than on the `swap_local_id`).
```
thomaseizinger: Good idea, I like that.

We could even generalize that to an `address_book` table where we store IPs of other peers.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,181 @@
...
+
+Note that the role could be factored out of the table, but we decided against it because it adds unnecessary complexity for now.
+
+What remains is a simple table for `swap` which contains only the information which all swaps have in common: `role`, `alpha_protocol`, `beta_protocol` and `counter_party_peer_id`.
```
thomaseizinger: Nice!

I had a hunch that this is what we will end up with, good to see it confirmed :)

0026-an-extensible-db-design.adoc
thomaseizinger: Not a big fan of the `_swap_details` name (details is such a generic term), why not just `han_bitcoin_protocol`?

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+In order to store details for each swap such as data we need for creating the HTLCs for each protocol we introduce `\{protocol\}_swap_details`, where `\{protocol\}` can be `han_bitcoin`, ``han_ethereum`, `herc20` or `halight` for now.
+Similar to the `Swaps` table, we add information which is only known at a later point of time into a sub-table.
+This resulted in quite an excess of sub-tables such as `han_bitcoin/ethereum_redeem_identity`, `han_bitcoin/ethereum_refund_identity`, etc.
+
+Some notable sub-tables:
+
+* `han_bitcoin/ethereum_swap_detail` contains data needed to create the HTLCs. We decided to give this table a generated primary key, rather than using `swap_local_id` as the primary key to allow same chain swaps.
+* `han_bitcoin_ethereum_redeem/refund_identity`: Since some identities are exchanged during announcement, the identities are kept in extension tables.
+
+One could optimize here and combine all ledger specific identities into an own table `han_bitcoin_identity` to hold all Bitcoin `refund` and `redeem` identities.
```
thomaseizinger: I would even go one step further and not have `han_bitcoin_identity` tables but store the identities in the `han_bitcoin_protocol` table right away.

This will require nullable fields but we will have to have dedicated structs anyway for the DB stuff, so we can easily handle `Option<T>` fields there and do some error handling when converting them into our internal model.

It seems a bit of an overkill to me to introduce all these extra tables, just because we learn one of the two identities at a slightly later point.

Given that most of the time, both identities are going to be present, the JOIN overhead and complexity in the model doesn't seem to be worth it.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+This would mean though that we would need an additional unique id in those sub-tables, i.e. we can't use the `fk` as `pk`.
```
thomaseizinger: I would prefer that in general:

Let the DB have its own PK that is auto-generated and make dedicated columns for foreign-keys and our local-swap-id.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+* `han_bitcoin/ethereum_ledger_state`: is a new table which we introduced in this database schema.
```
thomaseizinger: Can we, by any chance, have a rendering of the model that excludes these tables? I am not keen on thinking about them now because I would delay introducing them until a later point!

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+It records observed ledger events such as `fund`/`redeem`/`refund`-transaction.
+We extracted the states into `han_state` in order to reduce duplication in each table and not have an extra column for the event-type (`fund`/`redeem`/`refund`).
```
thomaseizinger: Given that each even may come with different data, my initial guess would be that we should have one table per event, with a FK relation to the protocol (i.e. a row in the `han_bitcoin_protocol` table).

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+This allows us to reuse these states among all `han_*` protocols.
```
thomaseizinger: I would argue against this, see here: https://github.com/comit-network/comit-rs/issues/2533

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/halight_swap_protocols.puml&fmt=svg[HalightProtocols]
+
+=== Swap State
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/swap_states.puml&fmt=svg[SwapStates]
+
+Additional to the specific `alpha` and `beta` ledger state (depicted in <<Protocols>>) a `swap` has two additional states:
+
+. the _communication_ state which depicts the state during the announcement phase, i.e. `sent`, `accepted`, `rejected`
+. the aggregated state of the swap, aggregation of the communication state, ledger states, plus an additional error state, i.e. `in_progress`, `swapped`, `not_swapped`, `internal_failure`
```
thomaseizinger: I wouldn't track this in the DB, it is redundant data and can always be derived from everything else in the DB.

thomaseizinger (commented):
Nice work @da-kami and @bonomat !

Some questions / ideas :)

0026-an-extensible-db-design.adoc
bonomat: This statement was meant a bit differently and `model-driven approach` are probably the wrong words :D 

Let me try to clarify what was meant:

Right now, code is law, everything what is possible in the code is possible and we try to allow to store everything in the DB e.g. all combination of ledgers/protocols. We can (as you state correctly) limit that functionality by what cnd can store in the DB. This would still be in scope of this domain model. 

What was meant was that it is not scope of this spike to derive from the database what cnd can do and what not, e.g. what ledger combinations cnd can support and what not. 
E.g. cnd starts up, loads all possible ledger combinations, what networks, etc... from the db and is ready to go. 
While this partly the case, e.g. we have a _static_ table for the supported protocols, it was out of scope for this work to follow this approach to an extend where we add a table for supported ledgers/networks/token contracts/hash function/... to

0026-an-extensible-db-design.adoc
bonomat: I think we did not explicitly consider this.
 
I think the proposed approach is more elegant as otherwise there is an explosion of tables, i.e for each new supported protocol you will need to add a bunch of new tables vs right now we have 1.

I'm not sure I can follow how this leads to a more "type-safe" solution. 
The way how I would have implemented it is something like this: 

```
let swap: HanBtcHanEthSwapDetail = swaps::table
         .inner_join(
han_bitcoin::table.on(han_bitcoin::local_swap_id.eq(swaps::local_swap_id)),)
         .inner_join(
han_ethereum::table.on(han_ethereum::local_swap_id.eq(swaps::local_swap_id)),
                    )
                    .select((swaps::id, swaps::local_swap_id))
                    .filter(swaps::alpha_protocol.eq("han_bitcoin"))
                    .filter(swaps:beta_protocol.eq("han_ethereum"))
            })
```
That's as type safe as it gets right? :)

0026-an-extensible-db-design.adoc
bonomat: That is really a good point and goes partly towards your first question: https://github.com/comit-network/spikes/pull/62/files#r416269116

You are right though, we can easily integrate this into `swap` as we don't load this from cnd anywhere. This was actually a result of the normalization and thinking about it now, this might be one step too far for our needs. 

Side note: 
Given the tendencies of the team to discuss protocol names (😬) it is more efficient to have only a single fk to the protocol in another table instead of having the the protocol names in here. This allows us to only update one entry in `protocol` if we change the name of a protocol vs updating all related entries in `swap`.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+* HErc20
+* HALight
+
+Protocols based on scriptless scripts are not in focus, but known constraints (e.g. no secret) are taken into consideration.
+
+=== Database normalization
+
+Database normalization is a process to organize a database into tables and columns.
+The main idea with this is that a table should be about a specific topic and only supporting topics included.
+If a table is not properly normalized and has data redundancy then it will not only eat up extra memory space but will also make it difficult to handle and update the database, without facing data loss.
```
D4nte:
```suggestion
If a table is not properly normalized and has data redundancy then it will not only eat up extra memory space but will also make it difficult to handle and update the database, without facing data corruption.
```
I think it's more about corruption (redundant info not being properly sync'd between tables) than loss. The loss part is about ensure all the necessary information is indeed in the DB.

0026-an-extensible-db-design.adoc
bonomat: Sounds good and I think it's similar to what Tobin proposed.

0026-an-extensible-db-design.adoc
bonomat: Good question. I don't know, I think she always derives it from the secret right?

0026-an-extensible-db-design.adoc
bonomat: That's what we do right? This table stores the known `address` to a `peer_id`. Or do you mean something different here?

0026-an-extensible-db-design.adoc
bonomat: You are probably right. We drove it to an extreme and aimed for zero null values :)

0026-an-extensible-db-design.adoc
bonomat: I would leave the model here for the sake of completeness but can add an excerpt of it without the sate? That'd be straight forward.

0026-an-extensible-db-design.adoc
bonomat: Oh right. we did not consider the additional data for the states but only considered the tx_ids.

0026-an-extensible-db-design.adoc
bonomat: Can you elaborate on this please? I don't see the relation to the linked ticket.

0026-an-extensible-db-design.adoc
bonomat: We had this discussion as well and thought it would be more efficient to store in the DB than always derive it :)

0026-an-extensible-db-design.adoc
thomaseizinger: Ah I get you!

> E.g. cnd starts up, loads all possible ledger combinations, what networks, etc... from the db and is ready to go.

If we do that, we can just ditch Rust and use JavaScript right away :)
I don't think it is a good idea to go down that route. At least a lot of things that we tried to do in past are going in the complete opposite direction and embed _more_ things in the type system and not less.

0026-an-extensible-db-design.adoc
da-kami: Yes, we always compute the secret for Alice (or extract it on Bob's side):

https://github.com/comit-network/comit-rs/blob/e6fc2daf2635cab3aff347045702f1e24d819c8d/cnd/src/swap_protocols/rfc003/alice/actions/generic_impl.rs#L77-L84

When loading `AcceptedSwap`s there is not secret (we just store `Request` and `Accept` messages atm):

https://github.com/comit-network/comit-rs/blob/7a5e2ab59fdf5e99ef545cbf780ab9254a8de876/cnd/src/load_swaps.rs#L8-L33

0026-an-extensible-db-design.adoc
da-kami: I think this is actually a bit more problematic. 
We cannot properly derive all `swap_states`, because of the `INTERNAL_FAILURE` state. This one is orthogonal to the other states. Hence, I think this should actually be tracked in the DB...?

0026-an-extensible-db-design.adoc
thomaseizinger: > That's as type safe as it gets right? :)

Not really because there is no relation between:

```
.inner_join(
han_bitcoin::table.on(han_bitcoin::local_swap_id.eq(swaps::local_swap_id))
```
and
```
.filter(swaps::alpha_protocol.eq("han_bitcoin"))
```

There is not enforcement that the data for a particular protocol is stored in the table that the `alpha_protocol` column in the swaps table suggests. Or am I missing something?

> I think the proposed approach is more elegant as otherwise there is an explosion of tables, i.e for each new supported protocol you will need to add a bunch of new tables vs right now we have 1.

Depends on how you define elegance I guess.
To me, an elegant design makes it impossible to express things that you want to be impossible when you run the program. It may result in more tables but they all really just have a couple of columns with foreign keys and no data is duplicated.

0026-an-extensible-db-design.adoc
thomaseizinger: Yeah, as mentioned above, I think these kind of names should be expressed in identifiers (like table names or Rust structs) and not as runtime data.

0026-an-extensible-db-design.adoc
thomaseizinger: She can definitely. My question was, did you intend to store it for Alice as-well?

0026-an-extensible-db-design.adoc
thomaseizinger: The emphasis of my comment was on the plural of IP :)

There could be several valid IPs for a single party and we might want to store them over longer periods of time.

0026-an-extensible-db-design.adoc
da-kami: We generally don't model the event properly, but mesh the event into the state and just track the tx ids. I wonder how far we want to take this. 
The idea is that the blockchain remains the base of truth, but the tx_id helps us to speed things up. (@bonomat correct me if I'm wrong :)

0026-an-extensible-db-design.adoc
thomaseizinger: Yes please, I would just want to look at it without the state (as I think we are not gonna go there yet?)

0026-an-extensible-db-design.adoc
thomaseizinger: The ticket suggests that there should be no abstract "han" concept which means no reuse between those protocols, at least not on a conceptual level. On an implementation detail level, we can decide to reuse logic that is the same but I don't think there is much to reuse apart from names :D

0026-an-extensible-db-design.adoc
thomaseizinger: For efficiency reasons, it might be helpful to separate completed swaps (where the status will never change again) from ones that are in progress.

As long as data is changing, I would hesitate to store things that can run out of sync.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+We identified the following data sources needed:
+
+. Initial data of each swap when being created for a specific role for a specific alpha and beta protocol where alpha and beta is one of: `HAN`, `HErc20`, `HALight`
+. Additional data added to the swap for a specific role (e.g. secret hash only known to Bob during announcement)
+. Additional data added to the swap's alpha or beta ledger data.
+
+While not necessarily in scope we considered the following as well:
+
+. Swap State (one per swap, aggregated state - record all states with timestamp)
+. Communication State (one per swap - record all states with timestamp)
```
D4nte: Not sure what you mean at this stage.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+We identified the following data sources needed:
+
+. Initial data of each swap when being created for a specific role for a specific alpha and beta protocol where alpha and beta is one of: `HAN`, `HErc20`, `HALight`
+. Additional data added to the swap for a specific role (e.g. secret hash only known to Bob during announcement)
+. Additional data added to the swap's alpha or beta ledger data.
+
+While not necessarily in scope we considered the following as well:
+
+. Swap State (one per swap, aggregated state - record all states with timestamp)
```
D4nte: If it's an aggregation looks like we already know it's redundant.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+Instead of starting from scratch, we start iterating over a slightly adapted table layout based on the one proposed by @bonomat
+https://3.basecamp.com/4403044/buckets/16118249/messages/2583518881#__recording_2600021433[in basecamp].
+
+The final model can be found at the end of this document.
+Our reasoning for creating certain table relations is summed up in the sections below.
+
+=== Swaps
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/swap.puml&fmt=svg[Swaps]
```
D4nte: `counterparty_peer_id` can be contracted to `peer_id`.
Expiry and amounts are missing and needed to resume a swap, let's continue to read...

D4nte: Also, `name` in protocol table is actually a combination of information.
Sometimes it's just the protocol (halight) sometimes it's protocol and ledger.
I'd be keen to not shorcut here and properly have:
- name
- ledger
- asset

What if we want to support halight for Litecoin? Han for Litecoin?

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+Our first challange was depicting `alpha` and `beta` protocols for the swap.
```
D4nte:
```suggestion
Our first challenge was depicting `alpha` and `beta` protocols for the swap.
```

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+Our first attempt was to not indicate this in the `swap` table but instead just use foreign key relationships.
+For that to work we would need to indicate in the child table whether it is `alpha` or `beta` (see below in `swap_protocol`).
+In order to reduce the amount of duplicated data we decided to add a separate table `protocol` and a `1 to many` relationship from `swap` to `protocol`.
```
D4nte:
```suggestion
In order to reduce the amount of duplicated data we decided to add a separate table `protocol` and a `1 to 2` relationship from `swap` to `protocol`.
```

0026-an-extensible-db-design.adoc
D4nte: Bob needs to store the secret hash.

0026-an-extensible-db-design.adoc
D4nte: > We could even generalize that to an `address_book` table where we store IPs of other peers.

What does rust-libp2p give us for that? I don't want to re-invente the advanced libp2p wheel here.

0026-an-extensible-db-design.adoc
da-kami: Yeah, that why we decided NOT to do that :)
`cnd` should depend on the database as little as possible; it is purely for recovery reasons.

0026-an-extensible-db-design.adoc
da-kami: At the moment: no. 
My reasoning would be, that we have another source of security assumptions if we do that. At the moment there is only the seed and the local swap id are used to compute the actual secret. If we explicitly save the secret then there is a third "source" for where the secret is coming from. I'd avoid that.

Of course you could argue that, if Alice changes her seed (for whatever reason) she would lose the ability to compute historic swaps' secrets. But those are stored on the blockchain anyway if redeemed, and if not redeemed it is irrelevant I'd say.

0026-an-extensible-db-design.adoc
da-kami: I see your point, this adds another layer of complexity though.
If we consider separating swaps that are `IN_PROGRESS` from `SWAPPED`/`NOT_SWAPPED`/`INTERNAL_FAILURE` ones, would we do that with different tables?
If so, would we then remove a swap that is is a final state(`SWAPPED`/`NOT_SWAPPED`/`INTERNAL_FAILURE`) from the `IN_PROGRESS` related table? (I suppose so.)

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+We added the `address_hint` extension table that relates to `swap` on the `counter_party_peer_id` field (rather than on the `swap_local_id`).
+
+Note that the role could be factored out of the table, but we decided against it because it adds unnecessary complexity for now.
+
+What remains is a simple table for `swap` which contains only the information which all swaps have in common: `role`, `alpha_protocol`, `beta_protocol` and `counter_party_peer_id`.
+
+=== Protocols
+
+[[HanProtocols]]
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/han_swap_protocols.puml&fmt=svg[HanProtocols]
```
D4nte: I am not convinced about the `han-bitcoin_swap_Detial_id` and other foreign keys, what not using the swap_local_id as a foreign and primary key?

D4nte: Why having the identities in separate tables?

D4nte: Why having the state in a separate table?

0026-an-extensible-db-design.adoc
D4nte: Why is it a separate table in the first place?

D4nte: Ah ok I see, the idea is to have the information related to the state in such table. Something that could be added to the diagram.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+=== Swap State
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/14e942f9c5ea1791f01bb65db0e09ac11cf5ab4c/assets/db_schema/swap_states.puml&fmt=svg[SwapStates]
+
+Additional to the specific `alpha` and `beta` ledger state (depicted in <<Protocols>>) a `swap` has two additional states:
+
+. the _communication_ state which depicts the state during the announcement phase, i.e. `sent`, `accepted`, `rejected`
+. the aggregated state of the swap, aggregation of the communication state, ledger states, plus an additional error state, i.e. `in_progress`, `swapped`, `not_swapped`, `internal_failure`
+
+Note, that there is one communication state and one aggregated swap state per swap, whereas the ledger-state is per alpha-/beta-ledger per swap.
```
D4nte: What is a communication state?

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,180 @@
...
+We decided that we want to track multiple communication and aggregated state events, thus we added the joint-tables `swap_communication_state_for_swap` and `swap_state_for_swap` to connect `swap` with the respective state table (`swap_communication_state`, `swap_state`.
+
+Note that all `*_state` tables are static tables similar to the `protocol` table.
+These tables contain fixed values that are only changed if the state-model changes.
+
+== Full DB Model
+
+[[COMITDBSchema]]
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/9e9f70785dd4281c808cc4c7ade8b2d14b133588/assets/0026-database-diagram.puml&fmt=svg[COMITDBSchema]
```
D4nte: What are `sent`, `accepted`, `rejected`?

0026-an-extensible-db-design.adoc
thomaseizinger: I gives us nothing to my knowledge, it is a library and doesn't interact with IO other than network communication.

0026-an-extensible-db-design.adoc
thomaseizinger: I would just avoid the overall concept of an aggregated state in the database. I think it goes against basic SQL DB rules where you don't store derived/redundant data.

You can simply have a `completed_at` column on the swap and if you check that one, you know you don't need to spawn watchers but can assume all the state is locally available and not gonna change.

0026-an-extensible-db-design.adoc
thomaseizinger: I think this is addressed by the solution to this ticket: https://github.com/comit-network/comit-rs/issues/2533

0026-an-extensible-db-design.adoc
thomaseizinger: events != state.

There is always just one state but multiple events.
If you want to model states, you need to think in `UPDATE` statements.
I think modelling events makes more sense here as they are the source of truth.

In any case, all of this is out of scope anyway. I should have never commented on it in the first place because it doesn't matter at the moment :sweat_smile:

0026-an-extensible-db-design.adoc
thomaseizinger: No one is talking about storing the secret?

But the answer to my question seems to be `no`, so I will consider this as resolved :)

0026-an-extensible-db-design.adoc
bonomat: The communication states of a swap are `sent`, `accepted`, `declined`.

0026-an-extensible-db-design.adoc
bonomat: > There is not enforcement that the data for a particular protocol is stored in the table that the alpha_protocol column in the swaps table suggests. Or am I missing something?

Do you mean that there is no fk relation between the alpha_protocol column and the table to find the swap_details? That is true indeed. We could add a fk in `{protocol}_swap_detail` to `protocol`. I'm not sure if this adds any more guarantees though as we need to ensure this when inserting the data and the table name depicts already what data should be stored in it. 

> Depends on how you define elegance I guess.
To me, an elegant design makes it impossible to express things that you want to be impossible when you run the program. It may result in more tables but they all really just have a couple of columns with foreign keys and no data is duplicated.

As far as I can see the type guarantees are the same for both approaches as in code we need to guarantee that the correct data is stored in the correct tables no?
I might be missing something here so please help me out :)

0026-an-extensible-db-design.adoc
bonomat: I don't think there is such a thing as a `1 to 2` relationship.

0026-an-extensible-db-design.adoc
bonomat: > There could be several valid IPs for a single party and we might want to store them over longer periods of time.

I didn't know this is possible, but yes, you are right, there can be multiple IPs to a single peerID.

0026-an-extensible-db-design.adoc
bonomat: > I am not convinced about the han-bitcoin_swap_Detial_id and other foreign keys, what not using the swap_local_id as a foreign and primary key?

It's mentioned a few lines below: 

```
* `han_bitcoin/ethereum_swap_detail` contains data needed to create the HTLCs. We decided to give this table a generated primary key, rather than using `swap_local_id` as the primary key to allow same chain swaps.
```

> Why having the identities in separate tables?

5 lines below: 

```
* `han_bitcoin_ethereum_redeem/refund_identity`: Since some identities are exchanged during announcement, the identities are kept in extension tables.
```

> Why having the state in a separate table?

another 2 lines further down: 

```
han_bitcoin/ethereum_ledger_state: is a new table which we introduced in this database schema. It records observed ledger events such as fund/redeem/refund-transaction. We extracted the states into han_state in order to reduce duplication in each table and not have an extra column for the event-type (fund/redeem/refund). This allows us to reuse these states among all han_* protocols.
```

0026-an-extensible-db-design.adoc
bonomat: I thought the definition of `han_*` protocols was that they all share the same 4 actions (fund/redeem/refund).

0026-an-extensible-db-design.adoc
bonomat: currently Alice `sends` a `swap_request` to Bob who can `accept` or `reject`. Lateron this will be `announce` and `finalized` (and I'm missing something else).

D4nte: You are describing `rfc003` behaviour, but the spike says that the following protocols are in scope:
```
* HAN
* HErc20
* HALight
```

0026-an-extensible-db-design.adoc
bonomat: on it.

0026-an-extensible-db-design.adoc
thomaseizinger: > As far as I can see the type guarantees are the same for both approaches as in code we need to guarantee that the correct data is stored in the correct tables no?

Without fk, you are free to store whatever "protocol" you want in the protocol columns and store the protocol data somewhere else, i.e. I can say "this swap is han-ethereum<->halight" in the swap table but actually store entries in the "han_bitcoin" and "herc20" protocol tables for this swap id.

Foreign-key relations in SQL (I hope in SQLite too) are _typed_, meaning a foreign key in a specific column can only point to one particular table and not any other.

Meaning if we create a schema like:

- `swap`
	- id
	- ...

- `han_ethereum_protocol`
	- id
	- wei_quantity
	- expiry
	- ...

- `han_bitcoin_protocol`
	- id
	- satoshi_quantity
	- expiry
	- ...

- `han_ethereum_han_bitcoin_swap`
	- id
	- alpha_ledger_fk -> `han_ethereum_protocol`
	- beta_ledger_fk -> `han_bitcoin_protocol`
	- swap_fk -> `swap`
	- ...

The database schema guarantees that entries in the `han_ethereum_han_bitcoin_swap` table always point to data of the actual protocols in use. In a way you avoid repeating yourself in saying what protocols are in use by just pointing to the relevant entries in the individual protocol tables. Due to the nature of typed FKs, this requires a dedicated table for each combination.

The database is going to prevent you from inserting invalid data because values in the `beta_ledger_fk` column MUST point to an entry in `han_bitcoin_protocol`, otherwise the SQL statement is not valid :)
You cannot make it point to a different protocol table (I hope that SQLite also enforces this, at least bigger database engines like PostgreSQL do that)

The downside of this is that we have one table per combination. They would only be join tables though to tie things together and not duplicate all the columns like we do in the current schema.

What we will need in any case (but that is independent of both solutions) is a way to determine the protocol combination of a swap with just the ID (in the code we currently call it `DetermineTypes`). This can be built on top of both solutions, with varying degrees of efficiency.

In summary, I am slightly preferring the solution with more safety (i.e. dedicated join tables) because that is what these tools are for. If we don't care about this kind of relational safety, then we are not getting as much out of SQL as we could and hence the argument of using it over a NoSQL solution becomes weaker.

Let me know what you think :)

0026-an-extensible-db-design.adoc
bonomat: @thomaseizinger : should be nicer now.

0026-an-extensible-db-design.adoc
D4nte: Ok cleared, it's created/finalized now.

0026-an-extensible-db-design.adoc
D4nte: Ok clarified, I was being picky on the possible values of the communications state.

0026-an-extensible-db-design.adoc
bonomat: I think I can mark this as resolved. Please reopen if you want to further this discussion.

0026-an-extensible-db-design.adoc
D4nte: True, my bad.

0026-an-extensible-db-design.adoc
D4nte: Cool thanks @thomaseizinger.

0026-an-extensible-db-design.adoc
D4nte: This has been addressed in another chain of comment by @thomaseizinger so marking this as resolved to not have two thread for the same topic.

0026-an-extensible-db-design.adoc
D4nte: Ok as discussed, the communication states will be `created` or `finalized`.

0026-an-extensible-db-design.adoc
bonomat: Awesome writeup. 

I think I get it, let me try to repeat: 

In the current proposal we have 

- `protocol`
	- id

- `swap`
	- id
	- alpha_protocol_fk: -> `protocol.id`
	- beta_protocol_fk: -> `protocol.id`
	- ...

- `han_ethereum_swap_details`
	- id
	- wei_quantity
	- expiry
	- ...

- `han_bitcoin_swap_details`
	- id
	- satoshi_quantity
	- expiry
	- ...

And since `han_ethereum_swap_details` and `han_bitcoin_swap_details` are not connected to the `protocol` we lose some type guarantees when wanting to load a particular swap.

I still think the type safetyness is the same despite the lack of this relationship due to how data is inserted. The database is going to ensure that the types in `han_ethereum_swap_details` and `han_bitcoin_swap_details` are of the correct type or better, the ORM will take care of this, e.g. we do `Swap.save()`, `HanBitcoinSwapDetails.save()` etc. And this is the important thing to ensure. 
So we need to ensure that the correct data will be stored in the code during insert time anyways. 
If we can't rely on this then in your proposal we would have a similar problem in a way that one can store multiple `{protocol}_swap_detail` combinations and wouldn't know which one to load (e.g. we store btc<>eth and ln_eth to with the same swap_id.

The reason why we introduced this `protocol` table and the `alpha/beta_protocol` column into `swap` was that we need to know what ledgers/protocols are involved in a swap by looking at the swap. 

I wanted to propose to just store the `protocol.id` in `{protocol}_swap_details` as well which should then fulfill your initial proposal but I realized that there is a problem with our proposal:
same chain swaps are not doable as we would not be able to identify which entry in `han_bitcoin_swap_detail` is alpha and which beta. 
Now, we can argue that this is out of scope anyways but we considered this during the design nevertheless which let to the current proposal. 

The last problem would be possible with your proposal by introducing a table `han_bitcoin_han_bitcoin_swap`.

0026-an-extensible-db-design.adoc
bonomat: I mark this as resolved. Please reopen and comment if you disagree.

0026-an-extensible-db-design.adoc
bonomat: > What if we want to support halight for Litecoin? Han for Litecoin?

In this case we would need to introduce a new table and change the nameing. I think once we have a solution from comit-network/comit-rs#2533 we can come back and update the db. For now we can just assume that `halight` is Bitcoin only. 

`Han` for Litecoin is possible with the proposed naming `Han_litecoin_swap_details` would be the new table.

0026-an-extensible-db-design.adoc
bonomat: I mark this as resolved and add a todo in the last comment.

0026-an-extensible-db-design.adoc
bonomat: I propose we hop on a call to discuss this further?

bonomat:
Open todos once the dust is settled: 

- [ ] every table should have an autogenerated id from the db
- [ ] existing fk used as pk can be declared as `UNIQUE`
- [ ] come up with a solution to the `swap`-`ledger`-`protocol`-problem from this comment https://github.com/comit-network/spikes/pull/62#discussion_r416270231

// to be extended

0026-an-extensible-db-design.adoc
thomaseizinger: > I propose we hop on a call to discuss this further?

Definitely, can you send out a calendar invite?

> If we can't rely on this then in your proposal we would have a similar problem in a way that one can store multiple {protocol}_swap_detail combinations and wouldn't know which one to load (e.g. we store btc<>eth and ln_eth to with the same swap_id.

Do you mean that two different join tables `han_ethereum_han_bitcoin_swap` and `han_bitcoin_herc20_swap` point to the same row in `swap`?

Agreed, this is a problem but I would argue that it is different one because it is not about messing up which protocols are involved with a particular swap but rather accidentally merging two swaps together. Each entry in those join tables is still consistent within themselves, it is just pointing to the wrong parent.
We could (partly) address that by not having a `swap` table and instead putting all the "common" fields into those tables (secret_hash, shared_swap_id, etc).

> The reason why we introduced this protocol table and the alpha/beta_protocol column into swap was that we need to know what ledgers/protocols are involved in a swap by looking at the swap.

As mentioned above, there are several ways of solving this. At the moment, we just [test which table contains an entry for a particular swap and return a static combination based on the table name](https://github.com/comit-network/comit-rs/blob/ba0cfe268f327d733d86f231803b315483faa37b/cnd/src/db/swap_types.rs#L26-L83).

> I still think the type safetyness is the same despite the lack of this relationship due to how data is inserted. The database is going to ensure that the types in han_ethereum_swap_details and han_bitcoin_swap_details are of the correct type or better, the ORM will take care of this, e.g. we do Swap.save(), HanBitcoinSwapDetails.save() etc. And this is the important thing to ensure.
So we need to ensure that the correct data will be stored in the code during insert time anyways.

I agree that the difference is minor, I still have an uneasy feeling that we rely on data that could potentially diverge to try and load the actual swap.
If the decision is to go with the current solution, I would at least like to have this documented :)

0026-an-extensible-db-design.adoc
da-kami: In my head it is like @bonomat says as well. Do we really want to get rid of the `HAN` abstraction ("protocol family") completely?

0026-an-extensible-db-design.adoc
da-kami: I am not sure that properly solves the problem, as there are different final states and it might be relevant which final state the swap had. 
Otherwise the final state `INTERNAL_FAILURE` would never be recorded. 
I think we would have to optimise the state mode, otherwise we should store the "derived" state - to my mind `INTERNAL_FAILURE` is actually not derived, or am I wrong?
We could save the final state when marking it complete. 🤔

0026-an-extensible-db-design.adoc
da-kami: My summary:

To add full safety we might really have to add a table for every combination of swap we allow, i.e. `han_btc_han_eth_swap`, `han_eth_han_btc_swap` (...), but that is not feasible. 

I feel adding the foreign relationship from `protocol` go `*_swap_detail` is a half baked solution, because nobody can stop you from putting a wrong protocol into a detail. (or differently put, this would be a column that always has the same entry...) - I feel we are modelling this "too human", assuming too many semantics.

I played around a bit with the DB tool we initially used for modelling, but could not come up with a better solution so far.
Somehow I feel the key could be to model `alpha` and `beta` more properly.

da-kami: I think modelling `alpha` and `beta` as joint tables for each ledger is a valid solution for this problem. 
This would grow linearly, we would have 2 tables more per ledger we add.

Consequences:
- the field `alpha_ledger` and `beta_ledger` in `swap` would be removed.
- The `protocol` table would be removed.

@thomaseizinger what do you think?

image

bonomat: With @da-kami's proposed solution it is very cumbersome to know what ledgers we are swapping in between.
If we are fine with this added complexity I would opt for a solution with less tables: Instead of having 2 more tables per `{protocol}_swap_detail` table I would opt for a de-normalized solution and just add 1 field to `{protocol}_swap_detail` indicating whether an entry is alpha or beta (and remove the protocol table and `alpha`/`beta` from `Swap`). 

this allows us to do two joins on Swap to load HanBTC and HanETH swap details: 
```
swap.left_join(han_btc where side=alpha).left_join(han_eth where side = beta)
```

thomaseizinger: Interesting idea, haven't considered that before!

That just stimulated another thought:

Nullable fk columns in the `swap` table:

- alpha_ledger_han_bitcoin -> `han_bitcoin_swap_details`
- beta_ledger_han_bitcoin -> `han_bitcoin_swap_details`
- alpha_ledger_han_ethereum -> `han_ethereum_swap_details`

etc

Have more than two of these columns set would be invalid and that could (in theory) be enforced through database triggers but let's not go there :sweat_smile: 

It would drastically reduce the number of tables AND we could determine with a single query, what the protocol combination of a swap is.

0026-an-extensible-db-design.adoc
da-kami: My opinion: Remove `swap_state` from the DB for now, it was not in scope in the first place.
I'd keep `ledger_state` and `communication_state` as it is.

0026-an-extensible-db-design.adoc
da-kami: > With @da-kami's proposed solution it is very cumbersome to know what ledgers we are swapping in between.
> If we are fine with this added complexity I would opt for a solution with less tables: Instead of having 2 more tables per `{protocol}_swap_detail` table I would opt for a de-normalized solution and just add 1 field to `{protocol}_swap_detail` indicating whether an entry is alpha or beta (and remove the protocol table and `alpha`/`beta` from `Swap`).
> 
> this allows us to do two joins on Swap to load HanBTC and HanETH swap details:
> 
> ```
> swap.left_join(han_btc where side=alpha).left_join(han_eth where side = beta)
> ```

@bonomat when we were working on the spike, the assumption was, that we would know up-front which combination we want to load. So that should not be a problem.

Your solution is fine for me too, and is definitely less complex in tables.

da-kami: I thought about all three ideas and came to the conclusion that all of them are valid solutions.
One thing that they have in common: The `protocol` table should go away. I think we agree on that. As @thomaseizinger mentions in a comment below, this level ob abstraction should be in "the table name" or column name for that matter.

Notes:
My solution builds purely on relationships between tables.

Thomas solution maps my solution from table-relationships to table-columns. We would have more and more columns that produce more and more null-values. 

Philipps solution is base on fields in the child rather than relationships - it solves the problem through value entries of `alpha` and `beta` - the relationship would be built on values.

Feel we covered all three levels of abstraction (table, column, field) 😄
For me it's hard to decide which one to go for. Mine is definitely the one with the highest complexity in joins 😬

0026-an-extensible-db-design.adoc
thomaseizinger: > I thought the definition of `han_*` protocols was that they all share the same 4 actions (fund/redeem/refund).

Yes that was the idea and I think you can still think of them like that if you want.
More interestingly though, it seems like that additional layer of abstraction is not worth it in many ways:

- In Rust code, it just leads to a large number of generics
- In the API, you always you need to look for the combination of `han` and `ethereum`/`bitcoin` anyway, otherwise you don't know how to semantically interpret the data.
- Same goes for the database.

It feels like this abstract protocol (iiee pfui, object-orientation) doesn't give us much because we hardly ever _abstract_ over it.
Interestingly, even though the actions are the same, the client doesn't really care about this because those actions are just "executable".

In other words: It doesn't feel like these protocols actually have something in common apart from using HTLCs (but that applies to all `H*` protocols). There is nothing special about the Ether HTLC that it **shares** with the Bitcoin one but differentiates it from the one for ERC20.

Doing an atomic swap on Bitcoin follows a completely different protocol than the Ethereum one does. We just invented shared terminology but that is about it :)

Conceptually, I am liking more and more the idea of cnd supporting a flat hierarchy of protocols where some are compatible with each other.

HErc20 and HAlight are the norm, HAN is the exception and should be aligned :)

0026-an-extensible-db-design.adoc
thomaseizinger: > My opinion: Remove `swap_state` from the DB for now, it was not in scope in the first place.
> I'd keep `ledger_state` and `communication_state` as it is.

Fine with me, but to make this explicit: You want to add tables now that we are not going to use initially?

0026-an-extensible-db-design.adoc
thomaseizinger: > @bonomat when we were working on the spike, the assumption was, that we would know up-front which combination we want to load. So that should not be a problem.

We have to load it from the database somehow. There is no other choice.

> Thomas solution maps my solution from table-relationships to table-columns. We would have more and more columns that produce more and more null-values.
> Philipps solution is base on fields in the child rather than relationships - it solves the problem through value entries of alpha and beta - the relationship would be built on values.

Let's go with Philipp's solution here. Seems to scale better. Nullable FK columns are kind of scary at the end of the day :D

I think with that, we've resolved all the comments! I just wanted to say good work everyone, I feel we have a solid understanding now and weighed off all the important aspects :)
I am really happy with the result!
Any chance we can get a short summary of this into the spike?

bonomat: Sounds good. I'll try to summarize this discussion in a short paragraph :)

bonomat:
Dear all reviewers and co-authors (@tcharding , @da-kami , @thomaseizinger , @D4nte ): I've updated the spike and integrated PR feedback :) 
I hope I haven't missed anything. 

Once ok, I'll squash the commits and update the links to use master.

0026-an-extensible-db-design.adoc
bonomat: I've split up `han_swap_details` to `han_bitcoin` and `han_ethereum`

0026-an-extensible-db-design.adoc
bonomat: I've removed `ledger_state` as this can be derived easily but kept `communication_state`.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,283 @@
...
+What remains is a simple table for `swap` which contains only the information which all swaps have in common: `local_swap_id` `role` and `counter_party_peer_id`.
+
+=== Protocols
+
+[[HanProtocols]]
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/86bfc97bcea91602eaf3e40e6da7d71beaff9911/assets/db_schema/han_bitcoin_swap_protocols_without_state.puml&fmt=svg[HanBitcoinProtocols]
+
+In order to store details for each swap such as data we need for creating the HTLCs for each protocol we introduce `\{protocol\}_swap_details`, where `\{protocol\}` can be `han_bitcoin`, ``han_ethereum`, `herc20` or `halight` for now.
+Similar to the `Swaps` table, we add information which is only known at a later point of time into a sub-table.
+This resulted in quite an excess of sub-tables such as `han_bitcoin/ethereum_redeem_identity`, `han_bitcoin/ethereum_refund_identity`, etc.
```
thomaseizinger: I think I commented somewhere that nullable columns for the identities might be a good idea as it gets rid of these extension tables. Did this get lost when you worked in the feedback or did you decide against it?

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,283 @@
...
+For completeness we depict `herc20` and `halight` diagrams below.
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/86bfc97bcea91602eaf3e40e6da7d71beaff9911/assets/db_schema/han_ethereum_swap_protocols_without_state.puml&fmt=svg[HanEthereumProtocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/86bfc97bcea91602eaf3e40e6da7d71beaff9911/assets/db_schema/herc20_swap_protocols_without_state.puml&fmt=svg[Herc20Protocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/86bfc97bcea91602eaf3e40e6da7d71beaff9911/assets/db_schema/halight_swap_protocols_without_state.puml&fmt=svg[HalightProtocols]
+
+
+==== Protocols with states:
```
thomaseizinger: Is your intention to actually implement these tables also right away or are they just here for inspiration once we want to have persistent state? Can we, for the sake of moving forward, exclude these for now? :)

I wouldn't want to discuss the details now unless you actually want to include those right away as-well.

0026-an-extensible-db-design.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/86bfc97bcea91602eaf3e40e6da7d71beaff9911/assets/db_schema/herc20_swap_protocols_with_state.puml&fmt=svg[Herc20Protocols]
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/86bfc97bcea91602eaf3e40e6da7d71beaff9911/assets/db_schema/halight_swap_protocols_with_state.puml&fmt=svg[HalightProtocols]
+
+=== Swap State
+
+image::http://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/comit-network/spikes/86bfc97bcea91602eaf3e40e6da7d71beaff9911/assets/db_schema/swap_states.puml&fmt=svg[SwapStates]
+
+Additional to the specific `alpha` and `beta` ledger state (depicted in <<Protocols>>) a `swap` has two additional states:
```
thomaseizinger: Same for these states. If you want to include them, I would have some more questions but I don't want to start the discussion if we agree on omitting them in the implementation for now.

assets/db_schema/all_entities.iuml
```diff
@@ -0,0 +1,343 @@
...
+secret_hash |o--|| swap
+
+@enduml
+
+@startuml(id=han_bitcoin_swap_protocols)
+
+entity "han_bitcoin_swap_detail" as han_bitcoin {
+    han_bitcoin_swap_detail_id : <<pk>>,
+    ---
+    swap_local_id : <<fk>>,
```
thomaseizinger: Only a minor thing, but can you actually create a foreign_key relationship on something that is not a primary key? I would have expected these to point to `swap.id`.

assets/db_schema/all_entities.iuml
```diff
@@ -0,0 +1,343 @@
...
+entity "shared_swap_id" as shared_swap_id {
+    id : <<pk>> <<fk>>,
+    ---
+    shared_swap_id
+}
+
+entity "address_hint" as address_hint {
+    id : <<pk>>
+    ---
+    counter_party_peer_id : <<unique>> <<fk>>,
```
thomaseizinger: What `fk` is this pointing to?

Couldn't this just be a table independent of swaps where we just keep track of address<->peerid mappings?

thomaseizinger (changes requested):
Looks good to me, only question that is important is about the identity extension tables :)

assets/db_schema/all_entities.iuml
da-kami: Yes, `swap.id` should be used for referencing inside the database! @bonomat we discussed this yesterday, the database should use the database internal identifier (the PK) for identifying the swap rather than the `swap_local_id`.

assets/db_schema/all_entities.iuml
da-kami: It points to `swap.counterparty_peer_id` (which us not a key as such, I was not sure at some point if that is actually a clean relationship).

The `peer_id` is mandatory to know for both parties for a swap.
So some relations has to exist between the swap and the `peer_id`.
We modelled it like this, to ensure that entering `peer_id`s would require a swap. But maybe that is actually not necessary. We can also keep the table independent of the `swap` and implement error handling if an `address_hint` is missing.

0026-an-extensible-db-design.adoc
bonomat: I didn't want to throw away the work, so they are just here for inspiration :)

assets/db_schema/all_entities.iuml
bonomat: Yes, this should be `swap.id`.

assets/db_schema/all_entities.iuml
bonomat: Yes, the `fk` should be removed. One can only link to primary keys or UNIQUE fields.

0026-an-extensible-db-design.adoc
bonomat: yes, this was feedback was lost 😇

bonomat: updated

assets/db_schema/all_entities.iuml
bonomat: I hope I caught all.

assets/db_schema/all_entities.iuml
bonomat: Not update :( 
I'll keep it locally instead of regenerating all links and will push it once approved.

thomaseizinger (approved):
Awesome work :clap: :100: :tada:

da-kami (commented):
Went over it, great work on getting all those changes worked in properly @bonomat!
I did not spot anything missing, let's go for merge! (can't `approve` because I am set as author ;)

tcharding (approved):

D4nte (approved):

PR #64, opened 2020-08-09
Add document: Standardizing COMIT Expiry Times

tcharding:
Complete re-write, thanks.

Currently we guess what expiry times to use for alpha and beta ledger. This document investigates how we can calculate useful expiries times.

### DIFF
--- a/0027-standardizing-comit-expiry-times.adoc
+++ b/0027-standardizing-comit-expiry-times.adoc
@@ -1,0 +1,133 @@
+= Standardizing COMIT Expiry Times
+Tobin C. Harding <tobin@coblox.tech>;
+:toc:
+:revdate: 2020-08-20
++ Created on: {revdate} +
+
+== Introduction
+
+In order for two parties to exchange assets (swap) using the COMIT protocol each party must take certain pre-defined actions at certain times.
+A swap involves two expiry times, one for each ledger (alpha and beta), we refer to these as alpha expiry and beta expiry respectively.
+What actions are available at what time is governed by the state of the swap and the expiry times.
+A swap is guaranteed to be atomic if, and only if, the correct sequence of actions is followed.
+
+It is therefore paramount that a user knows what stage the swap is up to and what step they should take next.
+
+The purpose of this document is to investigate two things
+
+* How should expiry times be calculated
+* Once calculated, how do these expiry times effect the actions of both parties in the swap as the swap progresses.
+
+== Useful Expiry Times
+
+On an abstract level, for a pair of expiries to be useful there exists three invariants:
+
+1. The beta expiry must elapse before the alpha expiry.
+2. Alice must have time to redeem the swap before the beta expiry elapses.
+3. Bob must have time to redeem the swap before the alpha expiry elapses.
+
+The first of these invariants does not change over time and will not be discussed further.
+
+Irrespective of how the expiry times are implemented, in order to reason about the expiry times, they can be thought of as relative to some fixed point which marks the start of the swap.
+For a swap to take place the expiries must be 'useful', useful is defined to mean a swap can successfully be completed with these expiries.
+Additionally, at any stage, each party should be able to answer the following question: "If I do the next step will these expiries still be useful for the counterparty?"
+
+== Modelling a Swap
+
+If we want to know if there is time enough to complete a swap we should know how long a swap takes.
+To predict how long a swap may take we should know a few things, what events must occur to complete the swap, in what order must they occur, do they overlap, and how long do they take.
+Therefore in order to reason about the expiry times in relation to the sequence of events that make up a swap we need to have some model of the swap.
+
+We aim to produce a state diagram that fully describes a successful atomic swap.
+
+Initial state diagram:
+
+        `Started` --> `Alpha Funded` --> `Beta Funded` --> `Beta Redeemed` --> `Alpha Redeemed` --> `Completed`
+
+There exists some sequence of actions that enables a swap to transition to 'Completed'.
+We first annotate the state diagram with actions and then describe each action below.
+
+Annotated state diagram:
+
+        `Started` - fund - mine - finality -> `Alpha Funded` - fund - mine - finality -> `Beta Funded` - redeem - mine - finality -> `Beta Redeemed` - redeem - mine - finality -> `Alpha Redeemed` --> `Completed`
+
+We now define the term 'done' to mean the swap is successfully completed from the perspective of the actor.
+For Alice 'done' is defined as `Beta Redeemed`.
+For Bob 'done' is defined as `Alpha Redeemed`.
+We also note that while the state transitions are the same for both parties the actions required to transition are not the same for both parties.
+
+For each action that an actor must take there is some reasonable time period in which the action can be completed.
+In order to reason about the expiries each actor must know the actions they must take but also the actions the counterparty must take.
+This is because we wish to reason about the usefulness of the expiries to both ourselves and our counterparty.
+At any stage of the swap the sum of the actions to transition from the current state to done represents the time required to complete the swap.
+
+We will now describe each of the actions that appears in the annotated state diagram.
+
+* fund, redeem: Fund and redeem refer to creating the fund/redeem transaction and submitting it to the blockchain network.
+* mine: The time it takes for an unconfirmed transaction to be mined i.e., included in a block.
+* finality: The time it takes for a transaction, once mined, to reach finality.
+
+Each of these times is specific to the actor, the transaction type (redeem/refund), and the underlying blockchain.
+
+We now model the swap from each actors perspective.
+
+Annotated state diagram by actor:
+
+Alice:
+
+        `Started` - fund - mine - finality -> `Alpha Funded` - fund - mine - finality -> `Beta Funded` - redeem - mine - finality -> `Beta Redeemed` --> `Done`
+
+Bob:
+
+        `Started` - fund - mine - finality -> `Alpha Funded` - fund - mine - finality -> `Beta Funded` - redeem(beta) - redeem(alpha) - mine - finality -> `Done`
+
+Here we see that as soon as the beta redeem transaction becomes known to the network Bob can create his redeem transaction and submit it to the alpha blockchain network.
+
+
+== Transition Period per Action
+
+=== Fund/Redeem
+
+There exists some amount of time that it takes for an actor to create a transaction and submit it to the network.
+Of consideration some transactions may be created in software some may take user intervention.
+A likely solution for this time period is defining some constant amount of time required by a user and some amount of time required by a software system.
+
+=== Mine
+
+The time it takes for a transaction to be include in the blockchain is determined by blocktime, network congestion, and transaction fees.
+We can dynamically determine a transaction fee with the aim of being included within some number of blocks.
+If we know the amount of time it takes to mine this many blocks then we can calculate a predicted time to mine.
+
+The amount of time it takes to mine N blocks is governed by the Poisson distribution of events.
+See https://en.wikipedia.org/wiki/Poisson_distribution for more information.
+
+=== Finality
+
+A transaction is said to reach finality when the probability of the block containing the transaction being immutable is above some threshold.
+Typically this is said to occur after some number of blocks has been mined on top of the block containing the transaction (referred to as number of confirmations).
+With a pre-defined value for the number of confirmations and the above mention Poisson distribution we can calculate the time it takes for a transaction to reach finality to within some threshold, finality is always probabilistic and never certain.
+
+== Notes on HTLC Expiry Times and Atomicity
+
+COMIT swaps are atomic due to the functionality provide by Hash Time Locked Contracts.
+Mapped to COMIT swap parlance, a user may not refund until after the expiry time has elapsed.
+
+In order for a swap to be guaranteed atomic the redeem transaction must be included in the blockchain _before_ the expiry time elapses.
+A transaction can not be said to be included until it has reached finality and is deemed immutable.
+After an expiry has elapsed a redeem transaction cannot be guaranteed because a refund transaction (i.e. another transaction spending the same outputs) could be submitted to the network at the same time.
+Given that the two transactions are mutually-exclusive, only one of them can be mined into a block.
+
+Therefore, there exists a time `T` after which attempting the redeem action is considered 'unsafe' because there is not enough time for the transaction to be mined and reach finality.
+
+== System Functionality
+
+The system is expected to provide the following functionality:
+
+.Requirements
+1. Given two blockchains, create a pair of useful expiries.
+2. Given a pair of expiries and their respective blockchains, determine if the expiries are useful.
+3. Given the current swap state, determine what the user should do.
+
+The expiries shall be represented as a duration i.e. a relative expiry time.
+Useful should be able to be calculated from some given reference point i.e., these expiries are useful if we want to start the swap at time T.
+Given the current swap state, some way of determining time from the perspective of the underlying blockchain, and the expiries the system should be able to determine whether the expiries are useful to both actors.
--- a/0027-standardizing-comit-expiry-times.adoc
+++ b/0027-standardizing-comit-expiry-times.adoc
@@ -89,7 +89,7 @@ Here we see that as soon as the beta redeem transaction becomes known to the net
 === Fund/Redeem
 
 There exists some amount of time that it takes for an actor to create a transaction and submit it to the network.
-Of consideration some transactions may be created in software some may take user intervention.
+Of consideration some transactions may be created in software some may take user interaction.
 A likely solution for this time period is defining some constant amount of time required by a user and some amount of time required by a software system.
 
 === Mine


0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+----
+TTI = N * <average block time>
+----
+
+=== Time To Final
+
+For each blockchain there exists a time, after a transaction has been included in a block, that the transaction is considered to be final.
+
+Confirmations represent the number of blocks in the block chain that have been accepted by the network since the block that includes the transaction. The number of confirmations a transaction has determines the probability that a blockchain fork will invalidate the transaction. With a given blockchain we define Time To Final as:
```
thomaseizinger:
```suggestion
Confirmations represent the number of blocks in the blockchain that have been accepted by the network since the block that includes the transaction. The number of confirmations a transaction has determines the probability that a blockchain fork will invalidate the transaction. With a given blockchain we define Time To Final as:
```

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+TTF = <number of required confirmations> * <average block time>
+----
+
+=== HTLC Expiry Time
+
+COMIT swaps are atomic due to the functionality provide by Hash Time Locked Contracts. Mapped to COMIT swap parlance, a user may not refund until after the expiry time has elapsed.
+
+The atomic guarantees of the COMIT protocols rely on the beta expiry time being _before_ the alpha expiry time, these expiry times are absolute (as opposed to relative) i.e., seconds since epoch.
+
+In order for a swap to be guaranteed atomic the redeem transaction must be included in the blockchain _before_ the expiry time elapses. Therefore the redeem action is subject to some `TTI` calculated for the given blockchain. Repeating because this is important; the expiry time represents a hard limit, the redeem transaction must be included in a block before the expiry elapses for the swap to be atomic.
```
thomaseizinger: If you fancy, you could go more into detail here of why it is not atomic otherwise: The expiry time represents the earliest moment in time where the refund transaction is considered to be valid and can be broadcasted to the network. Henceforth, starting from this point it becomes a race of which transaction is included first. Given that the two transactions are mutually-exclusive, only one of them can make it.

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+== Functionality
+
+The system is expected to provide the following functionality:
+
+.Requirements
+1. Given two blockchains, create a pair of useful expiries.
+2. Given a pair of expiries and their respective blockchains, determine if the expiries are useful.
+3. Given the current swap state, determine what the user should do.
+
+A pair of expiries is 'useful' if, at this moment, both parties are able to execute all steps to complete the swap.
```
thomaseizinger:
```suggestion
A pair of expiries is 'useful' if both parties are likely to execute all steps to complete the swap.
```

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+* Time to Act (`TTA`)
+* Time to Inclusion (`TTI`)
+* Time to Final (`TTF`)
+* HTLC Expiry time
+
+=== Time to Act
+
+For each actor there is a time frame within which they can be expected to act.
+
+Different users will have different response times e.g., a long running daemon is likely to respond to swap transitions more quickly than a user who is manually checking an application before doing each action. We define Time To Act (`TTA`) as the window of time within which the counterparty can expected to take the next action.
```
bonomat: >We define Time To Act (TTA) as the window of time within which the counterparty can expected to take the next action.

How do you justify the TTA?

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+=== Time To Inclusion
+
+For each blockchain there exists a time for a transaction to be included in a block.
+
+The likelihood of a transaction being included in a block is dependant on the transaction fee coupled with the current load on the blockchain network i.e., the number of current unconfirmed transactions. An increase in the transaction fee leads to an increase in the probability that the transaction will be included. Inclusion is not guaranteed, therefore the probability is defined to be the percentage chance of a transaction being included in a block within the next `N` blocks.
```
bonomat: > The likelihood of a transaction being included in a block is dependant on the transaction fee coupled with the current load on the blockchain network i.e., the number of current unconfirmed transactions.

The likelihood of a transaction being included also depends on the probability of a block being found. On average there is a block every 10 minutes but there is no guarantee that there will be a block every 10 minutes.

Good read on this: https://www.reddit.com/r/btc/comments/6v5ee7/block_times_and_probabilities/

bonomat: > Inclusion is not guaranteed, therefore the probability is defined to be the percentage chance of a transaction being included in a block within the next N blocks.

I don't see a percentage chance being used in your formula.

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+With a given fee and a given probability tolerance we define Time To Inclusion for a given blockchain as:
+
+----
+TTI = N * <average block time>
+----
+
+=== Time To Final
```
bonomat:
```suggestion
=== Time To Finality
```

Finality might be the term to use.

0027-safely-recommend-actions.adoc
bonomat: > The number of confirmations a transaction has determines the probability that a blockchain fork will invalidate the transaction.

A fork does not invalidate a transaction. A fork might result in the fact that the transaction never happened.

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+Note that step 2 is recursive, the base case being the final stage of the annotated state diagram - done (see below).
+
+=== Explanation
+
+Various events must occur in order to transition from one state to the next. We can identify these events and annotate the state diagram accordingly. If we are able to calculate the time of each of these events we are able to fully determine the period of each transition.
+
+We define the Time To Final to be a part of the transition to a state e.g., the alpha funded state is not reached until the alpha fund transaction has become final.
+
+There exists a transition diagram for each actor. Each actor progresses along their own state diagram but must be able to predict the state progression of the counterparty in order to determine whether progression is possible for the counterparty.
```
thomaseizinger: "Must predict" might be a bit too strong maybe? It has this co-notation of "we are fucked if that is not true" :grimacing: 

A way I like to think about it is that every actor (and hence every implementation of COMIT) needs to have a _model_ of what the counterparty is going to do. The model could be wrong and we should account for that.

This is effectively a way of ensuring that we build our stuff "trustless".

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+.Key:
+* (state): states are shown inside parenthesis
+* `TTA_A`: `TTA` for Alice
+* `TTA_B`: `TTA` for Bob
+* `TTI_AF`: `TTI` for Alpha Fund transaction
+* `TTI_BF`: `TTI` for Beta Fund transaction
+* `TTI_AR`: `TTI` for Alpha Redeem transaction
+* `TTI_BR`: `TTI` for Beta Redeem transaction
+* `TTF_A`: `TTF` for Alpha ledger
+* `TTF_B`: `TTF` for Beta ledger
```
thomaseizinger: Are these going to show up as identifiers in the code? :grimacing:

I am a bit worried that we are gonna have a soup of abbreviated variables that are very easy to mix-up if one makes good use of auto-complete given that they all start with the same two letters.

A tip for implementing this: assuming they would all be `u64`, re-using them would "just work" because they are all `Copy`. If you want to avoid that and let the compiler help you to not make a mistake, you might want to make newtypes that are purposely _not_ `Copy` so you get a "move" error when you re-use one of them accidentally.

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+.Note
+* `TTI_AR` is not actually essential for Bob, we assume Bob waits for Alice's redeem transaction to be included in a block, technically by monitoring unconfirmed transactions (e.g. watching the Bitcoin mempool) `TTI_AR` could be removed. This in no way effects the security of the protocol.
```
thomaseizinger: :+1:

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+
+
+== Calculating Events
+
+=== Time to Act
+
+Time To Act should allow for network congestion and expected response of the user.
+
+* `TTA_B` is defined to be 15 minutes
+* `TTA_A` is defined to be 60 minutes
```
thomaseizinger: This is based on the idea that Bob = nectar and therefore automated right?

Maybe we can just generalize this and say, an automated party is expected to respond within 15 minutes and a human within 60.

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+* *abort*: Take no further action (cancel before fund).
+
+The recommended next action can be fully determined from the following algorithm:
+
+----
+1. IF refund is available RETURN refund
+2. IF there is time to transition to next AND there is time for the subsequent action RETURN next
+3. IF have funded RETURN wait_for_refund
+4. RETURN abort
+----
```
bonomat: I find it hard to follow this algorithm. Especially the recursive part of step 2. Maybe a structured [pseudocode](https://en.wikipedia.org/wiki/Pseudocode#Example) approach would more suitable?

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+Various events must occur in order to transition from one state to the next. We can identify these events and annotate the state diagram accordingly. If we are able to calculate the time of each of these events we are able to fully determine the period of each transition.
+
+We define the Time To Final to be a part of the transition to a state e.g., the alpha funded state is not reached until the alpha fund transaction has become final.
+
+There exists a transition diagram for each actor. Each actor progresses along their own state diagram but must be able to predict the state progression of the counterparty in order to determine whether progression is possible for the counterparty.
+
+.Annotated state diagram for Alice:
+----
+(Create) TTA_A TTI_AF TTF_A (alpha funded) TTA_B TTI_BF TTF_B (beta funded) TTA_A TTI_BR TTC_B (beta redeemed) (done)
+----
```
bonomat: Sorry, I have no idea what that means :) If this is some kind of standard, can you link the source to the model please?

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+* P: probability ???
+* N: number of blocks ???
+
+=== Time To Final
+
+We consider a transaction final once `X` number of blocks have been accepted by the network since the transaction was included.
+
+.We define `X` to be
+* Bitcoin: 6
+* Ethereum: 12
```
thomaseizinger: This is not enough if it should be similar to Bitcoin :)

6 BTC blocks are effectively one hour, Ethereum mines 210 blocks in an hour :D

0027-safely-recommend-actions.adoc
bonomat: > * `TTI_AR` is not actually essential for Bob, we assume Bob waits for Alice's redeem transaction to be included in a block, technically by monitoring unconfirmed transactions (e.g. watching the Bitcoin mempool) `TTI_AR` could be removed. This in no way effects the security of the protocol.

Adding a note on why this is not essential could help the reader.

0027-safely-recommend-actions.adoc
```diff
@@ -0,0 +1,173 @@
...
+=== Time To Inclusion
+
+For each blockchain there exists a time for a transaction to be included in a block.
+
+The likelihood of a transaction being included in a block is dependant on the transaction fee coupled with the current load on the blockchain network i.e., the number of current unconfirmed transactions. An increase in the transaction fee leads to an increase in the probability that the transaction will be included. Inclusion is not guaranteed, therefore the probability is defined to be the percentage chance of a transaction being included in a block within the next `N` blocks.
+
+With a given fee and a given probability tolerance we define Time To Inclusion for a given blockchain as:
+
+----
+TTI = N * <average block time>
```
bonomat: We should mention that N depends on the amount, i.e. I would ask for at least 6 confirmations for $1mio transaction vs. 1 confirmation is fine for a $1000 transaction.

thomaseizinger (approved):
Thanks for this effort @tcharding !

Looks correct from my perspective :)

A stylistic note: Splitting sentences to one per line makes suggestions and editing using Git a bit easier.

Some nitpicks in naming:

1. The SRA terminology is not quite on point. In the phrase "Safely recommend actions", _safely_ is an adverb to the verb _recommend_. However, we have never had an issue where the act of recommending was "unsafe". Instead, executing the recommended action might be unsafe for the user. That however is not expressed using this phrase.
2. The document is titled "COMIT Protocol Actions" which is a bit generic given that we are not actually defining actions and diverges from what you wrote in your PR description. Personally, I would probably call the spike: "Standardizing COMIT expiry times". Standards can easily be implemented in software which is our goal here. Ultimately, that will lead to us being able to "recommend safe actions".

In summary, I think we can just drop the "safe actions" terminology. As pointed out in (2), the actions being safe is actually just a side-effect and specific to our in-house way-of-thinking of how we implement atomic swaps. The actual problem that is being solved here are "useful expiries". I know that the project is titled around "safe actions" but I think that just wasn't a particularly good name :sweat_smile: 

You might also wanna take a look at our previous spike on this topic: https://github.com/comit-network/spikes/blob/master/0001-basic-expiry-model.md

bonomat (commented):
Thanks for writing this up. I think you are on the right track. 

## Content 

I find the way of how TTA/TTI/TTF are expressed a bit too simplified and I would expect that TTI and TTF are functions over different parameters, e.g. 

TTI depends on 
* paid fees
* probability of blocks being found

Once we have the formula we can simplify and say that the probability of a block being found equals to the average block time. 
In case you haven't seen it: Lucas/Lloyd did a similar spike some time ago: 
* https://www.overleaf.com/project/5c89bcf2ee9fc951b7a82d9d
* https://github.com/comit-network/spikes/blob/master/0001-basic-expiry-model.md (afaik content outdated and links to the overleaf doc above)

## Structure & writing style
I haven't reviewed the writing style per se - let me know if you want that. One thing up front: please follow the `on-line-per-sentence` approach. This makes reviewing way easier. 

### Regarding the structure

I think the structure of the document can be improved as right now the headings don't lead through the document. One proposal would from me - but feel free to come up with another one: 
* Introduction
* Motivation (the different times)
* Specification (functionality and solution)
* Examples (concrete examples for BTC and ETH)

0027-safely-recommend-actions.adoc
tcharding: Used as suggested.

0027-safely-recommend-actions.adoc
tcharding: I added your words to the doc.

0027-safely-recommend-actions.adoc
tcharding: Fixed, thanks.

0027-safely-recommend-actions.adoc
tcharding: Cool.

0027-safely-recommend-actions.adoc
tcharding: > Are these going to show up as identifiers in the code?

These codes were just my way of showing the there is some amount of time required to transition that should be accounted for. _How_ they are calculated is only roughly outlined because, as you and @bonomat both point out I believe, calculating these things is heuristic at best.

0027-safely-recommend-actions.adoc
tcharding: Nice idea, incorporated. Thanks.

0027-safely-recommend-actions.adoc
tcharding: Is there a reason it should it be similar to Bitcoin? I just did a web search on what number of confirmations folks in the Ethereum ecosystem use and came up with 12.

tcharding:
> Thanks for writing this up. I think you are on the right track.
> 
> ## Content
> I find the way of how TTA/TTI/TTF are expressed a bit too simplified and I would expect that TTI and TTF are functions over different parameters, e.g.
> 
> TTI depends on
> 
> * paid fees
> * probability of blocks being found

This is exactly what they are meant to express, I must not have made that clear enough. I added the following two sentences 
```
`TTA`,`TTI`, and `TTF` all represent functions over _some_ parameters.
These acronyms are used as placeholders for those functions without fully defining how the functions will be defined.
```

> Once we have the formula we can simplify and say that the probability of a block being found equals to the average block time.
> In case you haven't seen it: Lucas/Lloyd did a similar spike some time ago:
> 
> * [overleaf.com/project/5c89bcf2ee9fc951b7a82d9d](https://www.overleaf.com/project/5c89bcf2ee9fc951b7a82d9d)
> * [`master`/0001-basic-expiry-model.md](https://github.com/comit-network/spikes/blob/master/0001-basic-expiry-model.md) (afaik content outdated and links to the overleaf doc above)

Thanks, I had not looked at this. There usage of math language and symbols is clearly more advanced than my own but I believe, at least in my head, I have included everything that they discuss.
 
> ## Structure & writing style
> I haven't reviewed the writing style per se - let me know if you want that. One thing up front: please follow the `on-line-per-sentence` approach. This makes reviewing way easier.

Yes, cheers, I forgot about this convention.

> ### Regarding the structure
> I think the structure of the document can be improved as right now the headings don't lead through the document. One proposal would from me - but feel free to come up with another one:
> 
> * Introduction
> * Motivation (the different times)
> * Specification (functionality and solution)
> * Examples (concrete examples for BTC and ETH)

I already spent two and a half days working on this document, I'm not going to invest any more time to re-write it but I will definitely consider your comment when I embark on writing my next technical document.

Thanks for the review.

0027-safely-recommend-actions.adoc
tcharding: I made an educated guess. I specifically am not investing time into picking perfect values for this project, rather the focus is on making sure every value that would be needed to configure this is present.

0027-safely-recommend-actions.adoc
tcharding: Man, this link is awesome, I did not have even the slightest inclination of this. Thanks

tcharding: It was intentionally left out :)

0027-safely-recommend-actions.adoc
tcharding: I had that at first. One of my aims in this doc was to use the smallest word that fully described what I meant 'Time To Final' does this equally as well as 'Time To Finality' so I went with 'final'.

0027-safely-recommend-actions.adoc
tcharding: Terminology corrected, thanks.

0027-safely-recommend-actions.adoc
tcharding: I used
```
IF refund is available (implies fund has occurred) AND expiry time has elapsed
        RETURN refund

IF there is time to transition from the current state to the next state AND there is time to complete the next action
        RETURN next

IF we have funded
        RETURN wait_for_refund

RETURN abort
```

0027-safely-recommend-actions.adoc
tcharding: I don't really know how to make this clearer. The acronyms represent functions that return a time value, these state diagrams include the time values required to transition from one state to the next. If there is some standard for writing this I'm happy to learn it.

0027-safely-recommend-actions.adoc
tcharding: I thought this was the note :)

0027-safely-recommend-actions.adoc
tcharding: I don't think N depends on a concrete amount. It depends on the perceived value of the amount. I may think that $1000 is a lot of money and needs 20 confirmations on Bitcoin but you may consider $1000 to be pocket change and be happy with 1 confirmation. For me this is implied in the configuration of number of required confirmations.

Is anyone even going to read this spike again after the implementation is done? I doubt it, since it does not live with the code (as I originally hoped) it is just an artefact of the dev process to me. I can definitely add a section on further configuration considerations that folks may wish to use.

0027-safely-recommend-actions.adoc
bonomat: I proposed to use `Finality` because it has a special meaning in the blockchain space ([random source from google](https://academy.binance.com/glossary/finality)). Feel free to leave `Final` though.

0027-safely-recommend-actions.adoc
bonomat: > I don't think N depends on a concrete amount. It depends on the perceived value of the amount. I may think that $1000 is a lot of money and needs 20 confirmations on Bitcoin but you may consider $1000 to be pocket change and be happy with 1 confirmation. For me this is implied in the configuration of number of required confirmations.

The reason why the amounts matter is because _attacking_ a blockchain costs money. A 1h/6 blockchain reorg for Bitcoin currently costs ~$500k in hashing power which means if you have a $1mio tx it's worth to invest those $500k into a 51% attack and create a 6 block deep re-org to double spend your $1mio transaction. On Ethereum a 1h attack on the otherhand would only cost you about $300k. 
It's not worth to spend $500k on a reorg if you want to double spend a $1000 tx. 
This means, the finality for a transaction of a blockchain depends on the amount and the current hash rate cost. 

A nice table on cost: https://www.crypto51.app/
A recent example of a 51% attack: https://www.coindesk.com/ethereum-classic-suffers-second-51-attack-in-a-week

tcharding: Cool, I did not know about this. I don't really think that we need to consider transactions with value greater than $300k at the moment when we do not even have a mainnet solution. I would think that it should be noted somewhere in the docs though. Thanks for educating me. Would documentation of this consideration be enough for you?

0027-safely-recommend-actions.adoc
tcharding: Oh cool, again good knowledge sharing thanks. I will use 'finality'.

0027-standardizing-comit-expiry-times.adoc
```diff
@@ -0,0 +1,200 @@
...
+:toc:
+:revdate: 2020-08-07
++ Created on: {revdate} +
+
+== Introduction
+
+In order for two parties to exchange assets (swap) using the COMIT protocol each party must take certain pre-defined actions at certain times.
+A swap is guaranteed to be atomic if, and only if, the correct sequence of actions is followed.
+It is therefore paramount that a user knows what stage the swap is up to and what step they should take next.
+Additionally, at any stage, each party should be able to answer the following question: "If I do the next step is there time, for myself and the counterparty, to do all subsequent steps?".
```
D4nte:
```suggestion
Additionally, at any stage, each party should be able to answer the following question: "If I do the next step is there time, for myself and the counterparty, to do all subsequent steps and keep the swap atomic?".
```

0027-standardizing-comit-expiry-times.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+== Swap State Machine
+
+A successful COMIT atomic swap (happy path) may be expressed as a series of state transitions:
+
+        Created --> Alpha Funded --> Beta Funded --> Beta Redeemed --> Alpha Redeemed --> Done
```
D4nte:
```suggestion
        Started --> Alpha Funded --> Beta Funded --> Beta Redeemed --> Alpha Redeemed --> Done
```
Or `t0` or `Ready`. Not sure what is actually created, at this stage, nothing is created, only both parties have finished communicating the parameters.

0027-standardizing-comit-expiry-times.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+`TTA`,`TTI`, and `TTF` all represent functions over _some_ parameters.
+These acronyms are used as placeholders for those functions without fully defining how the functions will be defined.
+
+=== Time to Act
+
+For each actor there is a time frame within which they can be expected to act.
+
+Different users will have different response times e.g., a long running daemon is likely to respond to swap transitions more quickly than a user who is manually checking an application before doing each action.
+We define Time To Act (`TTA`) as the window of time within which the counterparty can be expected to take the next action.
```
D4nte: More importantly, this is where the party can leverage the option problem and decide or not to proceed with the swap. I think this should be mentioned.

0027-standardizing-comit-expiry-times.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+=== Time To Inclusion
```
D4nte:
```suggestion
=== Time To Mined
```
Not too keen on the word `inclusion`. The usual terminology is `confirmed` or `mined`.

I prefer `mined` because `confirmed` can lead to confusion on whether it means `1 confirmation` or `confirmed to finality`.

In https://bitcoinfees.earn.com/#delay they call it `delay` -> `Delay To Mine`.

D4nte:
There has already been a couple of thorough review on the documents. I feel like adding my input would just create more work for little benefit so I'll refrain commenting more than the comments I parked yesterday and published today.
If you are happy to spend more time on this PR, let me know and I can review further.

tcharding:
> There has already been a couple of thorough review on the documents. I feel like adding my input would just create more work for little benefit so I'll refrain commenting more than the comments I parked yesterday and published today.
> If you are happy to spend more time on this PR, let me know and I can review further.

Thanks man, I am going to put a bit more work into this one, I'll ping you when I have.

0027-standardizing-comit-expiry-times.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+Here we see that as soon as the beta redeem transaction becomes known to the network Bob can create his redeem transaction and submit it to the alpha blockchain network.
+
+
+== Transition Period per Action
+
+=== Fund/Redeem
+
+There exists some amount of time that it takes for an actor to create a transaction and submit it to the network.
+Of consideration some transactions may be created in software some may take user intervention.
```
thomaseizinger: This sentence has a bit of a weird structure to me.

Also, the word "intervention" for me has the co-notation of "preventing" something to happen. "Interaction" might be better.

thomaseizinger (approved):
Thanks for diving into this and writing it up so thoroughly :)

0027-standardizing-comit-expiry-times.adoc
tcharding: Yep, nice. 'interaction' is better. Thanks

